PR #3, opened 2017-03-29
Fixes for the API method

stoatally:
When attempting to get the API results they were returned in `$responseData[0]` instead of the expected `$responseData['results']`. I'm not sure why that would be, I'm running the latest (non-rc) release.

I also make sure to close the curl connection before leaving the method.


stoatally:
Closing this, will make a new PR.

PR #8, opened 2017-04-01
WIP: Use templates to render the page rather than echo

stoatally:
I've begun moving the site to use templates rather than just echoing out stuff all over the place.

- [x] Improve the publish form so that it displays the name being published
- [x] Move the 'Your image is on the way' message into the main template, it should replace the form.
- [x] Move error messages into the main template.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,6 @@
+{
+    "require": {
+        "php": ">=7.1.0",
+        "stoatally/dom": ">=0.1"
+    }
+}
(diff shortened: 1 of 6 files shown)


stoatally:
Yes, absolutely. Will be working on it again this evening.

stoatally:
Ok, that is done, obviously there's still a lot of crufty PHP in there.

kauffj:
Sorry this never got merged, @billbitt took over this project and decided to change frameworks (and languages!).

PR #15, opened 2017-05-26
adding web sockets (part 1)

billbitt:
I added websockets implementation (using socket.io) to cover the retrieval of claims at spee.ch/name and spee.ch/name/claim_id

still needed: web sockets implementation for the publish functionality


helpers/lbryApi.js
```diff
@@ -37,7 +37,35 @@ function orderTopClaims(claimsListArray){
...
+function getClaimWithUri(uri, resolve, reject){
+	console.log(">> making get request to lbry daemon")
+	axios.post('http://localhost:5279/lbryapi', {
+			method: "get",
+			params: { uri: uri }
+		}
+	).then(function (getUriResponse) {
+		console.log(">> 'get claim' success...");
+		//check to make sure the daemon didn't just time out
+		if (getUriResponse.data.result.error === "Timeout"){
```
kauffj: It's a good idea to define strings that have special values as constants.

helpers/lbryApi.js
```diff
@@ -37,7 +37,35 @@ function orderTopClaims(claimsListArray){
...
+			reject("get request to lbry daemon timed out");
+		}
+		console.log(">> response data:", getUriResponse.data);
+		console.log(">> dl path =", getUriResponse.data.result.download_path)
+		// resolve the promise with the download path for the claim we got
+		/* 
+			note: put in a check to make sure we do not resolve until the download is actually complete 
+		*/
+		resolve(getUriResponse.data.result.download_path);
```
kauffj: If you find yourself accessing the same field repeatedly, like getUriResponse.data, it can be good to just dereference it.

Additionally, assuming you're writing ES6 (which is probably a good idea), you can just write:

`.then(({data}) => {
  if (data.result.error === "Timeout")
  //etc.
})`

helpers/lbryApi.js
```diff
@@ -56,93 +84,81 @@ module.exports = {
...
-				method: "claim_list",
-				params: {
-					name: claimName
+
+	getClaimBasedOnNameOnly: function(claimName){
+		// 1. create a promise
+		var deferred = new Promise(function (resolve, reject){
+			// 2. code to resolve or reject the promise
+			// make a call to the daemon to get the claims list 
+			axios.post('http://localhost:5279/lbryapi', {  // receives a promise
```
kauffj: This should probably be defined as a constant.

PR #32, opened 2017-06-16
mysql - record and check state of local files

billbitt:
Added mysql table to track Files stored locally.  The serve routes now check the Files table before trying to perform a 'get' on the daemon.  If a local match is found based on claim_id, it checks the outpoint and either serves that local file or retrieves an updated one.

### DIFF
--- a/config/default.json
+++ b/config/default.json
@@ -7,6 +7,6 @@
 	},
 	"Database": {
 		"MySqlConnectionUri": "none",
-		"DownloadAddress": "C:\\lbry\\speech\\hosted_content"
+		"PublishUploadPath": "none"
 	}
 }
--- a/config/default.json
+++ b/config/default.json
@@ -4,5 +4,9 @@
 	},
 	"AnalyticsConfig":{
 		"googleId": "none"
+	},
+	"Database": {
+		"MySqlConnectionUri": "none",
+		"DownloadAddress": "C:\\lbry\\speech\\hosted_content"
 	}
 }
--- a/helpers/lbryApi.js
+++ b/helpers/lbryApi.js
@@ -91,7 +91,7 @@ module.exports = {
 				"params": { "uri": uri}
 			}).then(function(response){
 				console.log(">> 'resolve' success");
-				resolve(response.data);
+				resolve(response.data.result);
 			}).catch(function(error){
 				console.log(">> 'resolve' error");
 				reject(error);
--- a/helpers/lbryApi.js
+++ b/helpers/lbryApi.js
@@ -1,4 +1,5 @@
 var axios = require('axios');
+var db = require("../models");
 
 module.exports = {
 	publishClaim: function(publishParams){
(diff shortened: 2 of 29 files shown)


helpers/lbryApi.js
```diff
@@ -1,15 +1,27 @@
...
+				db.File.create({
+					name: publishParams.name,
+					claim_id: response.data.result.claim_id,
+					outpoint: response.data.result.outpoint,
+					file_name: "test",
+					file_path: publishParams.filePath,
+					file_type: fileType,
+					nsfw: nsfw,
+				}).catch(function(error){
+					console.log('An error occurred when writing to the MySQL database:', error);
```
kauffj: It might be a good idea to declare an extremely simple error logging class or function that simply calls `console.log` internally. That way when you want to replace it later with more sophisticated logging, you can do so easily, rather than needing to track down all the log calls.

Not a blocking suggestion.

helpers/lbryApi.js
```diff
@@ -23,21 +35,33 @@ module.exports = {
...
 				}
-				// resolve the promise with the download path for the claim we got
 				/* 
 					note: put in a check to make sure we do not resolve until the download is actually complete (response.data.completed === true)
 				*/
-				resolve(getResponse.data);
-			}).catch(function(getUriError){
+				// save a record of the file to the Files table
+				db.File.create({
+					name: response.data.result.name,
```
kauffj: This section can be simplified by avoiding repetition of `response.data.result`.

helpers/lbryApi.js
```diff
@@ -23,21 +35,33 @@ module.exports = {
...
+					console.log('An error occurred when writing to the MySQL database:', error);
+				});
+				// resolve the promise
+				resolve(response.data.result);
+			}).catch(function(error){
 				console.log(">> 'get' error");
 				// reject the promise with an error message
-				reject(getUriError);
+				reject(error);
 				return;
```
kauffj: (unnecessary return)

helpers/lbryHelpers.js
```diff
@@ -1,6 +1,7 @@
 var path = require('path');
```
kauffj: `lbryHelpers.js` sounds like and sort of looks like the kind of file that is on track to accumulate a lot of functions and become a mess. Worth looking at the typical patterns for unique or business-level modal functions in this framework.

helpers/lbryApi.js
```diff
@@ -1,15 +1,27 @@
 var axios = require('axios');
+var db = require("../models");
```
kauffj: Writing all paths as relative from the root project folder rather than relative from the current file will make for later easier refactoring IMO.

config/default.json
```diff
@@ -4,5 +4,9 @@
 	},
 	"AnalyticsConfig":{
 		"googleId": "none"
+	},
+	"Database": {
+		"MySqlConnectionUri": "none",
+		"DownloadAddress": "C:\\lbry\\speech\\hosted_content"
```
kauffj: What's this?

helpers/lbryApi.js
billbitt: :heavy_check_mark:

helpers/lbryApi.js
billbitt: :heavy_check_mark:

helpers/lbryHelpers.js
billbitt: may have gone a little overboard, but I did a big reorganize to handle this and hopefully lay the groundwork for better organization going forward.  Separated out true controller functions from helper methods.

helpers/lbryApi.js
billbitt: good point.  will aim to address on next update.

helpers/lbryApi.js
billbitt: agreed.  will wait and address this with the next big update which will be focused on logging.

config/default.json
billbitt: it's the path for the folder that the uploads will be uploaded to.  I renamed it to 'PublishUploadPath'

PR #41, opened 2017-06-27
Api post route

billbitt:
- create api/publish as a public API which can be used to publish images through spee.ch
- refactored the publish controller so it's function(s) can be used by both the socket.io publish route (home page) as well as the /api/publish POST route.


kauffj:
:+1: for updating the README in conjunction with this!

1) Can `claim` parameter be `name`? It sounds more like that's what it is.
2) Can `license` be restricted to licenses that spee.ch supports, and can those be enumerated? (or linked to the place in code that defines them)

billbitt:
@kauffj yes, will do!

PR #42, opened 2017-06-27
Front end and api

billbitt:
- changed 'claim' to 'name' in /api/publish 
- added format validation to /api/post submissions 
- updated readme
- added media queries so content will size and display dynamically to client window size
- removed unused bootstrap and MDB libraries


billbitt:
@kauffj made updates noted in last PR here

PR #78, opened 2017-07-13
Trending images 2

billbitt:
- associated the File and Request tables (Files have many Requests, and Requests belong to one File)
- for 'trending images', created one sql query that handles joining and grouping the requests rather than sorting the data in js

### DIFF
--- a/controllers/statsController.js
+++ b/controllers/statsController.js
@@ -148,59 +148,10 @@ module.exports = {
     logger.debug('retrieving trending requests');
     const deferred = new Promise((resolve, reject) => {
       // get the raw requests data
-      db.Request
-        .findAll({
-          where: {
-            createdAt: {
-              gt: startDate,
-            },
-            FileId: {
-              not: null,
-            },
-          },
-          include: [db.File],
-        })
-        .then(data => {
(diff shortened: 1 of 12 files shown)


controllers/statsController.js
```diff
@@ -138,61 +145,13 @@ module.exports = {
...
-            return b[0] - a[0];
-          });
-          sortedArray = sortableArray.map((a) => {
-            return a[1];
-          });
-          // return results
-          resolve(sortedArray);
+      // get the raw requests data
+      db.sequelize
+        .query('SELECT COUNT(*), file.* FROM request LEFT JOIN file ON request.FileId = file.id WHERE FileId != "null" AND nsfw != 1 GROUP BY FileId ORDER BY COUNT(*) DESC LIMIT 25;', { type: db.sequelize.QueryTypes.SELECT })
```
kauffj: you probably want `WHERE FileId IS NOT NULL`

billbitt: fixed https://github.com/lbryio/spee.ch/commit/5ea7d959b9d960ef10000885a9867df477a0daf0

PR #93, opened 2017-07-19
Short urls 2

billbitt:
- added functions to retrieve the full claimId from a given short url, and vice versa.
- updated 'serve' and 'show' functions to allow for short url usage 
- separated 'serve' and 'show' functions into separate controllers 
    - Note: these are very similar functions and a bit duplicative.  However, I separated them in order to improve performance on certain routes.  They can probably be refactored and potentially combined to some degree.
- removed meme-fodder leftovers


helpers/libraries/serveHelpers.js
```diff
@@ -29,39 +98,157 @@ module.exports = {
...
+        switch (filteredClaimsList.length) {
+          case 0:
+            reject(new Error('That is an invalid short url'));
+            break;
+          case 1:
+            resolve(filteredClaimsList[0].claim_id);
+            break;
+          default:
+            const sortedClaimsList = filteredClaimsList.sort((a, b) => {
+              return a.height > b.height;
```
kauffj: Technically two claims could have the same height, though this is rather unlikely.

PR #98, opened 2017-07-22
Accept spaces in filename, convert them to dashes

98farhan94:
Accept space in filename from user, and replace the space to dash. Since a lot of users are used to typing out with spaces, better UX and less errors thrown out for invalid names.

### DIFF
--- a/helpers/libraries/publishHelpers.js
+++ b/helpers/libraries/publishHelpers.js
@@ -26,10 +26,12 @@ module.exports = {
     }
 
     // validate name
-    const invalidCharacters = /[^A-Za-z0-9,-]/.exec(name);
+    const invalidCharacters = /[^A-Za-z0-9,- ]/.exec(name);
     if (invalidCharacters) {
       throw new Error('The name you provided is not allowed.  Please use A-Z, a-z, 0-9, and "-" only.');
     }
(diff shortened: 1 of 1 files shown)


helpers/libraries/publishHelpers.js
```diff
@@ -26,10 +26,12 @@ module.exports = {
     }
 
     // validate name
-    const invalidCharacters = /[^A-Za-z0-9,-]/.exec(name);
+    const invalidCharacters = /[^A-Za-z0-9,- ]/.exec(name);
     if (invalidCharacters) {
       throw new Error('The name you provided is not allowed.  Please use A-Z, a-z, 0-9, and "-" only.');
     }
+    name = str.replace(/\s+/g, '-'); //replace space with dash/hyphen
```
billbitt: Good suggestion, but the filter for invalid characters on line 29 to 32 will catch any spaces and will throw an error.  Also, this function is taking place server-side, and I would rather prompt the user to correct their submission on the client side as @filipnyquist suggested.  It's good to have this check server side in case someone publishes by posting to the API route and avoids the client interface, but we should do the validation client side where possible.  I have a PR ready to go which will handle that client side check.

PR #99, opened 2017-07-22
Client side name validation

billbitt:
- Revised some of the client side publish functions to make the flow more clear.  Added comments where appropriate
- Added validation of the file name to check for special characters
- Added an event listener to the name input box to change spaces to dashes when they are entered
- Changed the function that auto-fills the name input box so that it replaces spaces with dashes


public/assets/js/publishFunctions.js
```diff
@@ -22,28 +22,28 @@ function validateFile(file) {
 			}
 			break;
 		default:
-			throw new Error('The ' + file.Type + ' content type is not supported.  Only, .jpeg, .png, .gif, and .mp4 files are currently supported.')
+			throw new Error(file.type + ' is not supported a supported file type. Only, .jpeg, .png, .gif, and .mp4 files are currently supported.')
 	}
+	// validate the file name (note: different from the lbry claim name)
```
billbitt: @98farhan94 This function is used to validate the file on the client side before it is submitted.  I added a check to make sure there are no special characters in the file name.  Unfortunately, the file.name in the file object is read only so we can't simply replace the spaces.  For now we will just prompt the user to fix their file name.

public/assets/js/publishFunctions.js
```diff
@@ -22,28 +22,28 @@ function validateFile(file) {
...
-	var invalidCharacters = /[^A-Za-z0-9,-]/.exec(name);
-	if (invalidCharacters) {
-		throw new NameError(invalidCharacters + ' is not allowed. A-Z, a-z, 0-9, and "-" only.');
-	} else if (name.length < 1) {
-		throw new NameError("You must enter a name for your claim");
-	}
-}
-
+// validation function that checks to make sure the claim name is not already claimed
 function validateClaimName (name) {
```
billbitt: @98farhan94 note that there are two functions in this file.  One to validate the file itself, and one to validate the claim name that the user chooses.  The file name can have spaces in it, but the claim name cannot.

public/assets/js/publishFunctions.js
```diff
@@ -65,41 +65,87 @@ function validateClaimName (name) {
 	});
 	return deferred;
 }
+// validation function which checks all aspects of the publish submission
```
billbitt: @98farhan94 this function is run when the submit button is clicked to submit the file for publishing.

public/assets/js/publishFunctions.js
```diff
@@ -65,41 +65,87 @@ function validateClaimName (name) {
...
 		previewReader.onloadend = function () {
 			dropzone.style.display = 'none';
 			previewHolder.style.display = 'block';
-			previewHolder.innerHTML = '<img width="100%" src="' + previewReader.result + '" alt="image preview"/>';
-			
+			previewHolder.innerHTML = '<img width="100%" src="' + previewReader.result + '" alt="image preview"/>';			
 		};
 	}
 	// set the name input value to the image name if none is set yet
 	if (nameInput.value === "") {
```
billbitt: @98farhan94 this is where the claim-name input box is filled based off of the file name.  I added a check here to replace spaces with dashes as you suggested (line 123)

views/index.handlebars
```diff
@@ -10,7 +10,58 @@
...
+		// build new publish area
+		publishResults = '<p>Your publish is complete! You are being redirected to it now.</p>';
+		publishResults += '<p><a target="_blank" href="' + showUrl + '">If you do not get redirected, click here.</a></p>';
+		// update publish area
+		document.getElementById('publish-active-area').innerHTML = publishResults;
+		window.location.href = showUrl;
+	});
+	
+	//event listener to filter claim name inputs
+	document.getElementById('publish-name').addEventListener('input', function() {
```
billbitt: @98farhan94 I added an event listener to the claim name input box so that if a space is entered it will be changed to a dash.  I suppose we could also use this filter to disallow all the other extraneous characters from being inputted.

public/assets/js/publishFunctions.js
98farhan94: Thanks - and also for all the comments, they are really helpful!

public/assets/js/publishFunctions.js
```diff
@@ -22,28 +22,28 @@ function validateFile(file) {
 			}
 			break;
 		default:
-			throw new Error('The ' + file.Type + ' content type is not supported.  Only, .jpeg, .png, .gif, and .mp4 files are currently supported.')
+			throw new Error(file.type + ' is not supported a supported file type. Only, .jpeg, .png, .gif, and .mp4 files are currently supported.')
 	}
+	// validate the file name (note: different from the lbry claim name)
+	var invalidCharacter = /[^\w.-\s()]/g.exec(file.name);
+	if (invalidCharacter) {
+		throw new Error('Special characters, such as "' + invalidCharacter + '", are not allowed in the file name.');
```
kauffj: Can this just be fixed/stripped rather than erroring?

public/assets/js/publishFunctions.js
```diff
@@ -22,28 +22,28 @@ function validateFile(file) {
...
-
+// validation function that checks to make sure the claim name is not already claimed
 function validateClaimName (name) {
 	var deferred = new Promise(function(resolve, reject) {
+		// validate the characters in the 'name' field
+		if (name.length < 1) {
+			reject(new NameError("You must enter a name for your claim"));
+			return;
+		}
+		var invalidCharacters = /[^A-Za-z0-9,-]/g.exec(name);
```
kauffj: Might want to consider adopting a validator pattern/standard. Look at validator patterns in a popular JS web framework, for example.

public/assets/js/publishFunctions.js
billbitt: Yes, but not easily at the moment.  1. I can't strip or fix it client side because the file name is read only on the file object.  I did some snooping on google and stack overflow and it seemed like it was more or less immutable.  2. The solution I imagine is then to rename the file when it is saved on the server.  Unfortunately, the socket.io uploader library I am using doesn't have a built in option to rename the file.  So, I think to fix would be to rename the file on the server side with the node filesystem tools.  I will make an issue to add this.

public/assets/js/publishFunctions.js
billbitt: I did a little bit of research on this.  There are some validator libraries out there, or at least we can make a more standard pattern that fits our unique needs.  Will make an issue.

PR #105, opened 2017-07-24
Add magic numbers

caseyparker:
Use the actual file rather than its name to validate type. I probably didn't do this in the best way, so if somebody can do better with the same idea that'd be great.


lyoshenka:
Closing for now

PR #108, opened 2017-07-25
File extensions

billbitt:
- show routes are now naked urls such as `spee.ch/example` or `spee.ch/example/2`
- serve routes are now urls with file extensions, such as `spee.ch/example.jpg` or `spee.ch/example/2.jpg`
- made a few minor cosmetic changes, including help at top right and linking the content freedom flag to the relevant wiki

### DIFF
--- a/controllers/showController.js
+++ b/controllers/showController.js
@@ -28,6 +28,7 @@ module.exports = {
               // 3. if a matching record is found locally, serve it
               if (result) {
                 // return the data for the file to be served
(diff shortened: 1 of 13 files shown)


kauffj:
This is progress. I noticed it basically ignores the file extension, which I like.

However, `spee.ch/example.jpg` and `spee.ch/example/2.jpg` returning the same asset feels really wrong to me. We need to think of another structure that retains `example.jpg` as the last element.

PR #115, opened 2017-07-26
updated file limit 10mb for images

billbitt:
- increase image file size
- fixed typo


billbitt:
reddit user advised higher upload size and found typo

PR #160, opened 2017-09-09
Proposed solution to isue 159 toggleSection in Firefox

noliverio:
Firefox does not recognize the proprietary property 'event.srcElement'. The option to use event.target was added.


billbitt:
thanks @noliverio !

PR #162, opened 2017-09-11
Use existing thumbs for channel content

billbitt:
- Added thumbnails for app-published videos viewed via the `spee.ch/@channelname` channel url
- Added thumnails for app-published videos to opengraph/twitter meta tags

I need to refactor how the `Request` table records to associate them with records in the `Claims` table in order to extend thumbnail support to the spee.ch/popular listing.


controllers/serveController.js
```diff
@@ -192,6 +195,11 @@ module.exports = {
             return db.resolveClaim(fileInfo.name, fileInfo.claimId);
           })
           .then(resolveResult => {
+            if (!resolveResult.thumbnail || resolveResult.thumbnail === '') {
+              fileInfo['thumbnail'] = 'https://spee.ch/assets/img/content-freedom-large.png';
```
kauffj: DRY violation ;)

public/assets/css/componentStyle.css
```diff
@@ -154,27 +154,50 @@ button.copy-button {
 	height: 1em;
```
kauffj: We ought to consider adopting a CSS class naming standard. I'm a fan of BEM http://getbem.com

This is non-blocking, but should probably be done before a large HTML/CSS change (such as the re-design).

routes/page-routes.js
```diff
@@ -50,15 +50,14 @@ module.exports = (app) => {
...
-  // route to display all free public claims at a given name
+  // route to send embedable video player (for twitter)
   app.get('/embed/:claimId/:name', ({ params }, res) => {
     const claimId = params.claimId;
     const name = params.name;
-    const dummyParam = '.b';
     console.log('claimId ==', claimId);
     console.log('name ==', name);
     // get and render the content
-    res.status(200).render('embed', { layout: 'embed', claimId, name, dummyParam });
```
kauffj: :clinking_glasses:

public/assets/css/componentStyle.css
billbitt: agreed

controllers/serveController.js
billbitt: updated

billbitt:
tested on production server.  merging.

PR #167, opened 2017-09-12
Local store publishes

billbitt:
- user's publishes are stored in local storage, and displayed below the publish tool on the home page
- updated the css for links site-wide.  Still need to do a BEM overhaul

### DIFF
--- a/views/index.handlebars
+++ b/views/index.handlebars
@@ -3,6 +3,7 @@
 	<div class="full"> 
 		{{> publish}}
 		{{> learnMore}}
+        {{> recentPublishes}}
 	</div>
 	{{> footer}}
 </div>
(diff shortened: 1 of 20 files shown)


views/index.handlebars
```diff
@@ -50,17 +51,33 @@
...
-		var showUrl = msg.name + '/' + msg.result.claim_id;
+		var showUrl = msg.result.claim_id + '/' + msg.name;
 		// build new publish area
 		publishResults = '<p>Your publish is complete! You are being redirected to it now.</p>';
-		publishResults += '<p><a target="_blank" href="' + showUrl + '">If you do not get redirected, click here.</a></p>';
+		publishResults += '<p><a class="link--primary" target="_blank" href="' + showUrl + '">If you do not get redirected, click here.</a></p>';
 		// update publish area
 		document.getElementById('publish-active-area').innerHTML = publishResults;
+		// store the publish in local storage
+        const myStorage = window.localStorage;
```
kauffj: Storing this data server side and using a key to recover it would give you more flexibility in the data structure.

As-is, if this data structure changes, we'd have to either write client-side migrations or invalidate all client-side data.

One option is to store a token unique to publisher (user) on each entry in the file table. Knowing this token could grant edit rights in addition to acting as a retrieval filter.

views/partials/recentPublishes.handlebars
```diff
@@ -0,0 +1,59 @@
...
+
+<script type="text/javascript" >
+    const publishDisplayArea = document.getElementById('recent-publishes');
+    let myPublishes = null;
+    function addToPublishList(publishData, key, publishList){
+        let listElement = document.createElement('li');
+        listElement.setAttribute('id', key);
+        let link = document.createElement('a');
+        link.setAttribute('href', `/${publishData.claimId}/${publishData.name}`);
+        link.setAttribute('class', 'link--primary');
```
kauffj: Does this have to be done this way? Looks like another benefit of storing this server side and using a token -- we can use our actual templating engine!

billbitt:
will be replaced by PR #170

PR #170, opened 2017-09-21
Authentication

billbitt:
PR in progress (almost done, currently testing).


config/default.json
```diff
@@ -1,6 +1,7 @@
 {
 	"WalletConfig": {
-		"LbryClaimAddress": "none"
+		"LbryClaimAddress": "none",
+		"DefaultChannel": "none"
```
kauffj: Great design here! Love that this makes it easy to use different channels on staging/dev and production as well as easy to change if forked.

config/default.json
```diff
@@ -11,6 +12,8 @@
 		"Password": "none"
 	},
 	"Logging": {
-		"LogLevel": "none"
+		"LogLevel": "none",
+		"SlackErrorChannel": "none",
```
kauffj: Is there a reason this file uses `"none"` as a string instead of `null` or something falsier?

controllers/publishController.js
```diff
@@ -7,23 +7,27 @@ module.exports = {
...
       // 1. make sure the name is available
-      publishHelpers.checkNameAvailability(publishParams.name)
+      publishHelpers.checkClaimNameAvailability(publishParams.name)
       // 2. publish the file
       .then(result => {
         if (result === true) {
           return lbryApi.publishClaim(publishParams);
         } else {
-          return new Error('That name has already been claimed by spee.ch.  Please choose a new claim name.');
+          return new Error('That name has already been claimed by spee.ch.');
```
kauffj: "This name is already in use by spee.ch"

"Claim" is an unusual word that is very LBRY specific. Let's avoid using it until we have to.

controllers/publishController.js
```diff
@@ -7,23 +7,27 @@ module.exports = {
   publish (publishParams, fileName, fileType) {
     return new Promise((resolve, reject) => {
       let publishResults = {};
+      let file;
+      let claim;
       // 1. make sure the name is available
-      publishHelpers.checkNameAvailability(publishParams.name)
+      publishHelpers.checkClaimNameAvailability(publishParams.name)
```
kauffj: Does the daemon support two publishes to the same name if made from different channels?

models/user.js
```diff
@@ -0,0 +1,34 @@
+module.exports = (sequelize, { STRING }) => {
+  const User = sequelize.define(
+    'User',
```
kauffj: Is this a "User" or a "Channel"? I understand these have a 1-1 correspondence right now, but these are likely not the same thing as spee.ch continues to grow/evolve.

I think it is likely that this should be two tables, a User table and a Channel table, with Channel having a foreign key to User.

In addition to being more semantic, this would facilitate growth into a design where one user can access multiple channels from a single session/login without a tricky later refactor.

helpers/lbryApi.js
```diff
@@ -120,4 +122,22 @@ module.exports = {
         });
     });
   },
+  createChannel (name, claimAddress) {
+    return new Promise((resolve, reject) => {
+      axios
+        .post('http://localhost:5279/lbryapi', {
```
kauffj: Should be a const somewhere

helpers/lbryApi.js
```diff
@@ -120,4 +122,22 @@ module.exports = {
...
+          method: 'channel_new',
+          params: {
+            channel_name: name,
+            amount      : 0.1,
```
kauffj: Should also probably be a const, assuming this is the same value for channels and publishes.

helpers/publishHelpers.js
```diff
@@ -72,7 +73,7 @@ module.exports = {
     if (title === '' || title === null) {
       title = name;
     }
-    if (description === '' || title === null) {
+    if (description === ' ' || description === null) {
```
kauffj: `description.trim() === ''`

controllers/publishController.js
```diff
@@ -36,14 +40,34 @@ module.exports = {
...
           claimId: publishResults.claim_id,
         };
-        return Promise.all([db.upsert(db.File, fileRecord, upsertCriteria, 'File'), db.upsert(db.Claim, fileRecord, upsertCriteria, 'Claim')]);
+        // create the records
+        return Promise.all([db.upsert(db.File, fileRecord, upsertCriteria, 'File'), db.upsert(db.Claim, claimRecord, upsertCriteria, 'Claim')]);
       })
-      .then(() => {
+      .then(result => {
+        file = result[0];
+        claim = result[1];
```
kauffj: `[file, claim] = result`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

helpers/handlebarsHelpers.js
```diff
@@ -0,0 +1,96 @@
...
+const config = require('config');
+
+module.exports = {
+    // define any extra helpers you may need
+  googleAnalytics () {
+    const googleApiKey = config.get('AnalyticsConfig.GoogleId');
+    return new Handlebars.SafeString(
+            `<script>
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
```
kauffj: Why do these have to be in a helper instead of just in a normal template?

(I know nothing about Handlebars -- you can just say it has to be this way if that's the case :) )

public/assets/css/generalStyle.css
```diff
@@ -123,12 +146,40 @@ table {
...
+	color: green;
+}
+
+.info-message--failure {
 	color: red;
-	font-size: small;
 }
 
+input:-webkit-autofill {
+ -webkit-box-shadow: 0 0 0px 1000px white inset;
```
kauffj: What if I'm not on Chrome?

public/assets/js/authFunctions.js
```diff
@@ -0,0 +1,28 @@
+function sendAuthRequest (channelName, password, url) {  // url === /signup or /login
+    return new Promise(function(resolve, reject) {
+        // make sure the claim name is still available
+        let xhttp;
+        const params = `username=${channelName}&password=${password}`;
+        console.log(params, url);
+        xhttp = new XMLHttpRequest();
```
kauffj: Is this the only place we're making XMLHttpRequests in the app? If so, I suppose this is okay, if not, we might want to standardize this a bit.

views/partials/channelCreation.handlebars
```diff
@@ -0,0 +1,51 @@
...
+     const password = document.getElementById('new-channel-password').value;
+     const channelNameErrorDisplayElement = document.getElementById('input-error-channel-name');
+     const passwordErrorDisplayElement = document.getElementById('input-error-channel-password');
+     const wrapper = document.getElementById('channel-publish-wrapper');
+     // prevent default so this script can handle submission
+     event.preventDefault();
+     // validate submission
+     validateNewChannelSubmission(channelName, password)
+         .then(() => {
+             wrapper.innerHTML = '<p>Creating your new channel...</p>';
```
kauffj: Updating HTML via JS can quickly become a maintainability nightmare.

IMO a better design is to render all the elements and use JS to toggle hiding/showing, either by acting directly on the elements or changing the class name of the parent and using CSS to dictate what is shown.

config/default.json
billbitt: :+1:

config/default.json
billbitt: I did it so that when I called the variable later, if l logged the variable and got 'none' I knew the variable had been created but wasn't updated with the proper value for the environment. Not very clean/efficient.   I fixed this now. They are all set to `null` and a script runs through them when the server loads and prints their values so I can check them up front. (https://github.com/lbryio/spee.ch/pull/170/commits/250ead165bd51c4b9812d514c724d320f526d90c)

controllers/publishController.js
billbitt: no, it will overwrite the first claim with the second, and just update the channel

controllers/publishController.js
billbitt: :+1:

models/user.js
billbitt: per our conversation today, I am updating the db to include both User and Channel tables

helpers/publishHelpers.js
billbitt: :thumbs:

helpers/handlebarsHelpers.js
billbitt: They have to be in a helper so that they can be generated based on the proper google analytics key from the environmental vars

billbitt:
I still have yet to abstract the lbryApi into a proxy design, but the rest of the issues have been address.  I'd like to commit this so that I can start working on the CSS.

PR #186, opened 2017-10-06
replace with nicer no claim page

MSFTserver:
image
speech-unclaimed
https://spee.ch/2c181a7a55cf5aea42a74945fde91844c6a6c8ab/speech-unclaimed.jpg


billbitt:
I fixed the existing no claims page.  thanks!

PR #203, opened 2017-10-17
fix share link alignment

MSFTserver:


MSFTserver:
need to find a better solution the text runs off the screen on smaller devices

PR #236, opened 2017-11-02
Slightly better wording (from/to)

Piste:
> Spee.ch is a media-hosting site that reads from and publishes content to the LBRY blockchain.

... reads slightly smoother than... 

> Spee.ch is a media-hosting site that reads and publishes content from the LBRY blockchain


kauffj:
Thanks @Piste! If you'd like a small LBC tip for this, please see https://lbry.io/faq/tips

PR #249, opened 2017-11-08
Db datatypes migration

billbitt:
- added migrations to change some data types in the db
- updated the config files


billbitt:
Updated to use DECIMAL

PR #250, opened 2017-11-08
general.css Video Skin Class Stlye Improvement

Smurphonomics:
Previously, the global video element was styled to make it look better on the stand alone page. However; the class `.showasset-light` is the video skin class assigned to the stand alone page only.

Based on this, I have optimised the style on this class and removed parts no longer needed from the `video-player` class, thereby not affecting any other pages for areas only required on the stand alone page.

### DIFF
--- a/public/assets/css/general.css
+++ b/public/assets/css/general.css
@@ -520,14 +520,13 @@ table {
    	display: block;
    	margin: 0 auto;
     	background-color: #fff;
+	cursor: pointer;
 }
-#video-player {
-    	cursor: pointer;
+#showlite-body #video-player {
   	margin-top: 2%;    	
     	padding: 6px;   	
     	max-width: 50%;
 	border: 1px solid #d0d0d0;
-	box-shadow: 0 0 50px -6px rgba(0, 0, 0, 0.72);
 }
 .showlite-asset {
     	max-width: 100%;
--- a/public/assets/css/general.css
+++ b/public/assets/css/general.css
@@ -517,22 +517,20 @@ table {
 	width: 100%;
 }
 #video-player, .showlite-asset {
-    padding: 6px;
-    border: 1px solid #555;
-    background-color: #fff;
+   	display: block;
+   	margin: 0 auto;
+    	background-color: #fff;
 }
 #video-player {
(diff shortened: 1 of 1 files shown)


public/assets/css/general.css
```diff
@@ -516,18 +516,23 @@ table {
...
-    padding:6px;
 }
 
 .showlite-asset {
     max-width: 50%;
+    margin: 0 auto;
+    margin-top: 2%;
+    padding: 6px;
+    border: 1px solid #a2a2a2;
+//Optional Box Shadow - Looks better on a darker than white background.
```
billbitt: I'm not sure about the box shadow in general, but I do think it works better on `#video-player` than on regular `showlite-asset` assets (e.g. pictures).  How about moving it to #video-player.

public/assets/css/general.css
```diff
@@ -516,18 +516,23 @@ table {
...
-    border: 1px solid #d0d0d0;
-    padding:6px;
 }
 
 .showlite-asset {
     max-width: 50%;
+    margin: 0 auto;
+    margin-top: 2%;
+    padding: 6px;
+    border: 1px solid #a2a2a2;
```
billbitt: I think I prefer the images without a border., but also the `margin: 0 auto;` doesn't seem to be taking affect.  I'd like to see what it looks like if the images are centered like the video.

Smurphonomics:
Yes. The above CSS was for video only. Does .showasset-light affect images too?

All images need are positions, such as `center`, `center `and maybe scaling to fit in the page?

I really need to set this app up so I can test locally. Will do this today!

Smurphonomics:
ID's need to be applied. What are the ID's for the following 2 pages?

https://spee.ch/b39388bb9ffa8946dde2acd49fbdc01561eb2edd/Speechless3
https://spee.ch/b/Speechless3.mp4

billbitt:
https://spee.ch/b39388bb9ffa8946dde2acd49fbdc01561eb2edd/Speechless3 - `<body id="show-body">`
https://spee.ch/b/Speechless3.mp4 - `<body id="showlite-body">`

PR #259, opened 2017-11-09
added id tags to body elements

billbitt:
@Smurphonomics what do you think?  I added an id to the body for each layout.

the layouts are as follows:

- main: general layout for homepage, about, login, 404, etc.
- channel: channel overview page
- show: layout for an asset displayed with details such as links and metadata
- showlite: layout for an asset served directly with no extra info/metadata

Is this going to be specific enough, or should we put an id on each view rather than at the layout level

### DIFF
--- a/views/layouts/channel.handlebars
+++ b/views/layouts/channel.handlebars
@@ -21,7 +21,7 @@
 	<!-- google analytics -->
 	{{ googleAnalytics }}
 </head>
-<body>
+<body id="channel-body">
     <script src="/assets/js/generalFunctions.js"></script>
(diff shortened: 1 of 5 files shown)


Smurphonomics:
Just as long as it's on each pages body tag, it'll be fine :)

Smurphonomics:
By that I mean all pages such as stand alone pages. Example, the page that contains the raw media such as...

https://spee.ch/930a27d8fb833148fac9ea0f183a3ab8ec02be80/LBRY-Space-Ball.jpg

PR #261, opened 2017-11-09
Undefined og description

billbitt:
updated controller to pass the required info (title and description) to the view renderer


controllers/publishController.js
```diff
@@ -67,7 +67,7 @@ module.exports = {
         resolve(publishResults); // resolve the promise with the result from lbryApi.publishClaim;
       })
       .catch(error => {
-        logger.error('publishController.publish, error', error);
+        logger.error('caught a publishController.publish error');
```
kauffj: Shouldn't the actual error be logged here?

If not, ignore.

helpers/handlebarsHelpers.js
```diff
@@ -17,56 +17,48 @@ module.exports = {
...
-    let basicTags = `<meta property="og:title" content="${title}"> 
-          <meta property="og:url" content="${showUrl}" > 
-          <meta property="og:site_name" content="Spee.ch" > 
-          <meta property="og:description" content="${description}">`;
+    console.log('title', title);
+    console.log('description', description);
+    let ogTitle = `<meta property="og:title" content="${title}" >`;
+    let ogUrl = `<meta property="og:url" content="${showUrl}" >`;
+    let ogSiteName = `<meta property="og:site_name" content="Spee.ch" >`;
+    let ogDescription = `<meta property="og:description" content="${description}" >`;
```
kauffj: Is there a fallback value for this when it is not set?

controllers/publishController.js
billbitt: The full error is passed back up the promise chain and then logged.  This particular log is more just for debugging, to be able to quickly ID if this error was triggered.

helpers/handlebarsHelpers.js
billbitt: no there isn't, and you are correct that should be added :)

PR #268, opened 2017-11-15
Drag and drop fix

billbitt:
- split up primary and preview dropzone functions
- created error display functionality for preview dropzone functions
- renamed cancelPublish function to more appropriate name

### DIFF
--- a/public/assets/js/dropzoneFunctions.js
+++ b/public/assets/js/dropzoneFunctions.js
@@ -1,53 +1,58 @@
+const dropzoneFunctions = {
+    drop_handler: function (event) {
+        event.preventDefault();
+        var dt = event.dataTransfer;
+        if (dt.items) {
+            if (dt.items[0].kind == 'file') {
+                var droppedFile = dt.items[0].getAsFile();
+                publishFileFunctions.previewAndStageFile(droppedFile);
+            }
+        }
+    },
+    dragover_handler: function (event) {
+        event.preventDefault();
+    },
+    dragend_handler: function (event) {
+        var dt = event.dataTransfer;
+        if (dt.items) {
+            for (var i = 0; i < dt.items.length; i++) {
+                dt.items.remove(i);
+            }
+        } else {
+            event.dataTransfer.clearData();
+        }
+    },
+    primary: {
+        drop_handler: function (event) {
+            this.dragexit_handler();
+            dropzoneFunctions.drop_handler(event);
+        },
+        dragenter_handler: function () {
+            var primaryDropzone = document.getElementById('primary-dropzone');
+            primaryDropzone.setAttribute('class', 'dropzone dropzone--drag-over row row--margined row--padded row--tall flex-container--column flex-container--center-center');
+            primaryDropzone.firstElementChild.setAttribute('class', 'hidden');
+            primaryDropzone.lastElementChild.setAttribute('class', '');
 
-function drop_handler(event) {
-    event.preventDefault();
-    // if dropped items aren't files, reject them
-    var dt = event.dataTransfer;
-    if (dt.items) {
-        if (dt.items[0].kind == 'file') {
-            var droppedFile = dt.items[0].getAsFile();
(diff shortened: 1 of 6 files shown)


public/assets/js/validationFunctions.js
```diff
@@ -7,7 +7,7 @@ const validationFunctions = {
         }
         if (/'/.test(file.name)) {
             console.log('file name had apostrophe in it');
-            throw new Error('apostrophes are not allowed in the file name');
+            throw new Error('Apostrophes are not allowed in the file name.');
```
kauffj: Rather than erroring, it might be preferable to just strip out the disallowed characters when a file is selected.

public/assets/js/dropzoneFunctions.js
```diff
@@ -1,53 +1,110 @@
...
+            document.getElementById('asset-preview').style.opacity = 1;
+        },
+        showError: function(errorMsg){
+            this.onmouseleave_handler();
+            // show error
+            const errorElement = document.getElementById('asset-preview-dropzone-error');
+            errorElement.setAttribute('class', 'flex-container--column flex-container--center-center position-absolute');
+            errorElement.innerHTML = `<p>${errorMsg}</p>`;
+            // set timer to hide error
+            setTimeout(this.hideError, 2000);
```
kauffj: While this is a bit of a corner case, this pattern can technically introduce a confusing situation where an error flashes extremely briefly. If I manage to generate 2 errors within the timeout window, the second error will be cleared by the first callback.

Additionally, I'm not a huge fan of clearing errors on a time-basis. It's generally preferable to clear the error once the user has taken an action that can be presumed to have corrected it (e.g. typing in the relevant field, selecting a new file, whatever).

(Neither of these necessarily have to be addressed to merge this, but you should at least be aware.)

public/assets/js/dropzoneFunctions.js
```diff
@@ -1,53 +1,110 @@
...
+                this.showError(error.message);
+                return;
+            }
+            publishFileFunctions.previewAndStageFile(selectedFile);
+        },
+        onmouseenter_handler: function () {
+            // show drag-and-drop instructions
+            document.getElementById('asset-preview-dropzone-instructions').setAttribute('class', 'flex-container--column flex-container--center-center position-absolute');
+            // make image lighter
+            document.getElementById('asset-preview').style.opacity = 0.2;
```
kauffj: Could this be done by `:hover` CSS pseudo-class?

kauffj: More generally, I would try to avoid direct style changes in Javascript and instead only perform class modifications. Modifying styles directly means your CSS no longer is sufficient to understand all of the styles inside of the app. If you instead use Javascript to only modify classes and use those classes to apply style changes via CSS, then your CSS still contains all information related to styling.

kauffj: (There may be a few exceptions to "only use classes", like hiding and showing.)

public/assets/js/dropzoneFunctions.js
```diff
@@ -1,53 +1,110 @@
...
-    }
-}
+    },
+    primary: {
+        drop_handler: function (event) {
+            event.preventDefault();
+            this.dragexit_handler();
+            var dt = event.dataTransfer;
+            if (dt.items) {
+                if (dt.items[0].kind == 'file') {
```
kauffj: If I drop something that is not a file, should I receive feedback on that? I'm not actually sure what you could drop that _isn't_ a file, so I'm not sure whether or not someone would be confused when they did so.

billbitt: tbh I got the pattern from mdn and wasn't sure: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop.
I looked into it, and other elements on the page are one item that can be dragged and dropped that isn't a file.  e.g. you can drag one of the navigation elements over the drop zone.  
I will add a catch for that so it updates the error message accordingly.

PR #269, opened 2017-11-16
Centers Images Vertically and Horizontally

Smurphonomics:
Centers images v and h for a more pleasing experience.

### DIFF
--- a/public/assets/css/general.css
+++ b/public/assets/css/general.css
@@ -530,6 +530,12 @@ table {
 }
 .showlite-asset {
     	max-width: 100%;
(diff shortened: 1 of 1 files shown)


Smurphonomics:
Yes I agree, that would also be better because the styles are already in general.css to use.

Probably best to add this to the video centring too not that the video page has it's own ID :)

I'm just testing it locally on all media elements to make sure it works as it should.

billbitt:
closing this PR because of above mentioned wanting to use flexboxes.  Also, when serving images directly to the browser, we want to stick with the imgur convention of displaying the image/gif in the top left.

PR #277, opened 2017-11-29
Show request refactor

billbitt:
- refactored serve/show routes
- show route returns show page immediately with content metadata
- show page updates user as it searches for content that is not held locally


billbitt:
changed base to `development` to follow GitFlow branching model

README.md
```diff
@@ -26,10 +26,10 @@ spee.ch is a single-serving site that reads and publishes images and videos to a
...
 * /api/resolve/:name
 	* example: `curl https://spee.ch/api/resolve/doitlive`
-* /api/claim_list/:name
-	* example: `curl https://spee.ch/api/claim_list/doitlive`
-* /api/isClaimAvailable/:name (returns `true`/`false` for whether a name is available through spee.ch)
-	* example: `curl https://spee.ch/api/isClaimAvailable/doitlive`
+* /api/claim-list/:name
+	* example: `curl https://spee.ch/api/claim-list/doitlive`
+* /api/is-claim-available/:name (returns `true`/`false` for whether a name is available through spee.ch)
+	* example: `curl https://spee.ch/api/is-claim-available/doitlive`
```
kauffj: Hooray capitalization consistency :100:

controllers/serveController.js
```diff
@@ -1,150 +1,43 @@
...
-  return claimInfo.thumbnail;
-}
+const NO_FILE = 'NO_FILE';
 
 module.exports = {
-  getAssetByClaim (claimName, claimId) {
-    logger.debug(`getAssetByClaim(${claimName}, ${claimId})`);
+  getClaimId (channelName, channelId, name, claimId) {
+    if (channelName) {
+      return module.exports.getClaimIdByChannel(channelName, channelId, name);
```
kauffj: `channelId` is still a claim id. I might call this something like `channelClaimId` to make that clearer.

controllers/serveController.js
```diff
@@ -1,150 +1,43 @@
...
-      db.Claim.getLongClaimId(claimName, claimId) // 1. get the long claim id
-        .then(result => {  // 2. get the asset using the long claim id
-          logger.debug('long claim id ===', result);
-          if (result === NO_CLAIM) {
-            logger.debug('resolving NO_CLAIM');
-            resolve(NO_CLAIM);
-            return;
-          }
-          resolve(getAssetByLongClaimId(result, claimName));
+      db.Claim.getLongClaimId(claimName, claimId) // get the long claim id
```
kauffj: No need to wrap `resolve` and `reject` if you are just passing them the same parameter. You can just do:

`db.Claim.getLongClaimId(claimName, claimId).then(resolve, reject)`

controllers/serveController.js
```diff
@@ -1,150 +1,43 @@
...
+          return db.Claim.getClaimIdByLongChannelId(result, claimName);  // 2. get the long claim id
         })
-        .then(result => { // 3. get the asset using the long claim id
-          logger.debug('asset claim id =', result);
-          if (result === NO_CHANNEL || result === NO_CLAIM) {
-            resolve(result);
-            return;
-          }
-          resolve(getAssetByLongClaimId(result, claimName));
+        .then(result => {
```
kauffj: (Same comment as above - unnecessary wrappers)

controllers/serveController.js
```diff
@@ -197,57 +90,47 @@ module.exports = {
...
+        return 'mp4';
       default:
-        logger.error('I did not recognize that method');
-        break;
+        logger.info('showing unknown file type as image/jpeg');
+        return 'jpeg';
+    }
+  },
+  chooseThumbnail (claimInfo, defaultThumbnail) {
+    if (!claimInfo.thumbnail || claimInfo.thumbnail.trim() === '') {
```
kauffj: I'd look at whether this trimming can be done in the model layer, rather than the controller (i.e. can we perform the TRIM in SQL or at least before the data makes it out of the model layer).

If you don't have to trim here, you can then just do things like:

`const thumb = claimInfo.thumbnail || DEFAULT_THUMBNAIL`

which is a lot cleaner and possibly obviates the need for the function entirely

helpers/handlebarsHelpers.js
```diff
@@ -13,10 +13,10 @@ module.exports = {
     return new Handlebars.SafeString(gaCode);
   },
   addOpenGraph (title, mimeType, showUrl, source, description, thumbnail) {
-    if (title === null || title.trim() === '') {
+    if (!title || title.trim() === '') {
```
kauffj: Same with this - let's sanitize data sooner. Ideally, the data could be sanitized in both cases _before_ it even makes it into the DB. Otherwise, you should clean it as close as you can to where it is retrieved. It'll make your code easier to write and will prevent introduction of bugs caused by forgetting to do a trim check.

helpers/sequelizeHelpers.js
```diff
@@ -1,23 +1,23 @@
 module.exports = {
-  returnShortId: function (result, longId) {
+  returnShortId: function (claimsArray, longId) {
```
kauffj: :+1: I love better names

views/layouts/showlite.handlebars
```diff
@@ -9,16 +9,16 @@
     <link rel="stylesheet" href="/assets/css/general.css" type="text/css">
     <link rel="stylesheet" href="/assets/css/mediaQueries.css" type="text/css">
 	<meta property="fb:app_id" content="1371961932852223">
-	{{#unless fileInfo.nsfw}}
```
kauffj: Is it possible for `showlite` and `show` (and possibly other layouts) to share more of their common logic (particularly `<head>`), inside of a shared template?

controllers/serveController.js
```diff
@@ -197,57 +90,47 @@ module.exports = {
...
-          })
-          .catch(error => {
-            logger.error('throwing serverOrShowAsset SHOW error...');
-            throw error;
-          });
+  getLocalFileRecord (claimId, name) {
+    return db.File.findOne({where: {claimId, name}})
+      .then(file => {
+        if (!file) {
+          return NO_FILE;
```
kauffj: Is there a reason we don't just use something like `null` to mean no record found? This goes for here and a few other places.

models/certificate.js
```diff
@@ -171,10 +171,27 @@ module.exports = (sequelize, { STRING, BOOLEAN, INTEGER, TEXT, DECIMAL }) => {
...
+      });
+    });
+  };
+
   Certificate.getLongChannelId = function (channelName, channelId) {
     logger.debug(`getLongChannelId(${channelName}, ${channelId})`);
     if (channelId && (channelId.length === 40)) {  // if a full channel id is provided
-      return new Promise((resolve, reject) => resolve(channelId));
+      return this.validateLongChannelId(channelName, channelId);
     } else if (channelId && channelId.length < 40) {  // if a short channel id is provided
```
kauffj: Since 40 is used as a magic/special value both inside of this function and elsewhere, it should be a constant somewhere. Probably on the `Claim` class?

models/index.js
```diff
@@ -69,6 +69,7 @@ db.upsert = (Model, values, condition, tableName) => {
     })
     .catch(function (error) {
       logger.error(`${tableName}.upsert error`, error);
+      throw error;
```
kauffj: This one seems important :)

public/assets/js/assetConstructor.js
```diff
@@ -0,0 +1,124 @@
+const Asset = function () {
+    this.state = {};
```
kauffj: I'd like to discuss this design and the purpose of it some in person. Please message me on chat and we'll get on a Hangout.

public/assets/js/assetConstructor.js
```diff
@@ -0,0 +1,124 @@
...
+                    that.showFailureMessage('Well this sucks, but we can\'t seem to phone home');
+                }
+            }
+        };
+        xhr.send();
+    };
+    this.getAsset = function() {
+        const that = this;
+        const uri = `/api/get-claim/${this.state.claimName}/${this.state.claimId}`;
+        const xhr = new XMLHttpRequest();
```
kauffj: This is fine, but it does look like for the browsers we're targeting that we can use `fetch`. This is a much nicer API than XHR. See: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

routes/api-routes.js
```diff
@@ -4,18 +4,35 @@ const config = require('../config/speechConfig.js');
...
 const { publish } = require('../controllers/publishController.js');
-const { getClaimList, resolveUri } = require('../helpers/lbryApi.js');
+const { getClaimList, resolveUri, getClaim } = require('../helpers/lbryApi.js');
 const { createPublishParams, validateApiPublishRequest, validatePublishSubmission, cleanseChannelName, checkClaimNameAvailability, checkChannelAvailability } = require('../helpers/publishHelpers.js');
 const errorHandlers = require('../helpers/errorHandlers.js');
-const { postToStats, sendGoogleAnalytics } = require('../controllers/statsController.js');
+const { postToStats } = require('../controllers/statsController.js');
 const { authenticateOrSkip } = require('../auth/authentication.js');
 
+function addGetResultsToFileData (fileInfo, getResult) {
```
kauffj: This is only used in one place? Possibly doesn't need to be a function? (Not sure, fine either way)

routes/serve-routes.js
```diff
@@ -1,39 +1,47 @@
...
-      return getAssetByChannel(channelName, channelId, name);
-    case CLAIM:
-      return getAssetByClaim(name, claimId);
-    default:
-      return new Error('that claim type was not found');
+function isUriAChannel (name) {
+  return (name.charAt(0) === CHANNEL_CHAR);
+}
+
+function returnChannelNameFromUri (uri) {
```
kauffj: You may want to look at using `lbryuri.js` for some of this: https://github.com/lbryio/lbry-app/blob/master/src/renderer/js/lbryuri.js

In particular, the `parse` and `build` functions (which I recommend just trying to call so you see what they do if reading the code is tricky).

routes/serve-routes.js
```diff
@@ -52,19 +63,21 @@ function extractPageFromClaims (claims, pageNumber) {
   return pageOfClaims;
 }
 
-function determineTotalPages (totalClaims) {
-  if (totalClaims === 0) {
+function determineTotalPages (claims) {
```
kauffj: Should all of this logic be in `routes` itself? (I don't actually know Express paradigms)

routes/serve-routes.js
```diff
@@ -81,178 +94,208 @@ function determineNextPage (totalPages, currentPage) {
...
+    .catch(error => {
+      handleRequestError('serve', originalUrl, ip, error, res);
+    });
+}
+
+function determineResponseType (uri, headers) {
+  let responseType;
+  if (uri.indexOf('.') !== -1) {
+    responseType = SERVE;
+    if (headers['accept'] && headers['accept'].split(',').includes('text/html')) {
```
kauffj: DRY (just put it in a function, it's fine if it's short)

views/partials/asset.handlebars
```diff
@@ -1,29 +1,37 @@
...
-        <img  class="image-show" src="/{{fileInfo.claimId}}/{{fileInfo.name}}.{{fileInfo.fileExt}}" />
-    </a>
-{{/ifConditional}}
\ No newline at end of file
+    </div>
+</div>
+
+<script type="text/javascript">
+
+    const asset = new Asset();
```
kauffj: Among the questions we'll discuss: why not just...
```
const asset = new Asset({
  src: "foo",
  claimId: "bar",
  ...
})
```

kauffj: Also, given that asset is used elsewhere in code, we might want this class to have a more specific name.

controllers/serveController.js
billbitt: I'm elevating this out of the model layer, but keeping at the controller level

routes/api-routes.js
billbitt: yeah, only used in one place at the moment (and perhaps something that should happen lower in the model layer), but I am trying to compartmentalize code into functions where possible for ease of reading as well as testing.

routes/serve-routes.js
billbitt: this was very helpful.  I made my own version since the uri on spee.ch is slightly different, but used this pattern.  Probably could still be cleaned up a bit, but it simplified the code a lot.

routes/serve-routes.js
billbitt: Agreed.  I moved it to the controller layer.

views/layouts/showlite.handlebars
billbitt: There may be more room for synergy, but for now I condensed the common header tags into a handlebars helper and tightened up the open graph/twitter tag helper.

PR #279, opened 2017-12-01
Update passwords

billbitt:
- added endpoint for updating password (`spee.ch/api/password`)
- cleaned up the use of passport.js


billbitt:
changed base to `development` to follow GitFlow branching model

kauffj:
LGTM though possibly shouldn't be at the root level in the project.

PR #300, opened 2017-12-14
Set logo link in navbar to use pointer cursor

JoshStrobl:
Setting the logo link to use cursor pointer keeps things consistent with the use of user agent styling setting cursor pointer on Upload, Popular, and About links. Makes it more obvious that it's actually clickable.

### DIFF
--- a/public/assets/css/general.css
+++ b/public/assets/css/general.css
@@ -466,6 +466,10 @@ table {
 	display: inline-block;
 }
 
+.nav-bar-logo {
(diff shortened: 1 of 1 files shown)


billbitt:
thanks @JoshStrobl ! message me in discord so I can send you a tip

JoshStrobl:
@billbitt Nah it was a super easy change, keep the LBC :smile: But thanks though!

PR #303, opened 2017-12-14
281 unfurl show links

billbitt:
- updated serve/show logic to show pages when requested by bots
- created a config file so site information like title and host could be set in one place and change based on environment (e.g. https://spee.ch for production, https://dev1.spee.ch for dev1).  This helps testing so links don't point to wrong server.

### DIFF
--- a/config/speechConfig.js.example
+++ b/config/speechConfig.js.example
@@ -23,9 +23,8 @@ module.exports = {
     uploadDirectory: null,  // enter file path to where uploads/publishes should be stored
   },
   site: {
-    name       : 'Spee.ch',
-    host       : 'https://spee.ch',
-    description: 'Decentralized video and content hosting.',
+    name: 'Spee.ch',
+    host: 'https://spee.ch',
   },
   publishing: {
     defaultTitle      : 'Spee.ch',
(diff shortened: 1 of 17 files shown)


config/speechConfig.js.example
```diff
@@ -22,4 +22,14 @@ module.exports = {
   files: {
     uploadDirectory: null,  // enter file path to where uploads/publishes should be stored
   },
+  site: {
+    name       : 'Spee.ch',
+    host       : 'https://spee.ch',
+    description: 'Decentralized video and content hosting.',
+  },
+  publishing: {
+    defaultTitle      : 'Spee.h',
```
kauffj: typo

config/speechConfig.js.example
```diff
@@ -22,4 +22,14 @@ module.exports = {
...
+    defaultThumbnail  : 'https://spee.ch/assets/img/video_thumb_default.png',
+    defaultDescription: 'Open-source, decentralized image and video sharing.',
```
kauffj: Where are both of these descriptions used?

billbitt: publishing.defaultDescription is used for open graph description on claims that do not have a description field.  stite.description was intended to be used wherever a description of the site itself was needed, but it turns out it wasn't needed so I removed it for simplicity (latest commit)

PR #314, opened 2017-12-21
304 truncate description

billbitt:
- moved open graph defaults from model layer to view layer
- added a check that truncates descriptions to 200 characters at most

### DIFF
--- a/helpers/handlebarsHelpers.js
+++ b/helpers/handlebarsHelpers.js
@@ -69,7 +69,9 @@ module.exports = {
     return new Handlebars.SafeString(gaCode);
   },
   addOpenGraph (claim) {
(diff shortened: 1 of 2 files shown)


helpers/handlebarsHelpers.js
```diff
@@ -1,5 +1,62 @@
...
+function determineOgThumbnailContentType (thumbnail) {
+  if (thumbnail) {
+    if (thumbnail.lastIndexOf('.') !== -1) {
+      return determineContentTypeFromFileExtension(thumbnail.substring(thumbnail.lastIndexOf('.')));
+    }
+  }
+  return '';
+}
+
+function createOpenGraphDataFromClaim (claim, defaultTitle, defaultDescription) {
```
kauffj: "Create" is possibly not the correct word here? It's also possible to just do:
```
return {
  foo: "val",
  bar: "val2",
}
```

PR #323, opened 2018-01-15
React/Redux - publish component

billbitt:
- converts the upload functionality on spee.ch's homepage into a react component
- uses redux to handle state for the publish component
- reusable subcomponents that can be shared across the application if/when spee.ch is converted to a true react app
- added webpack for bundling the react code


react/containers/PublishForm.jsx
```diff
@@ -0,0 +1,254 @@
...
+import PublishUrlInput from './PublishUrlInput.jsx';
+import PublishThumbnailInput from './PublishThumbnailInput.jsx';
+import PublishMetadataInputs from './PublishMetadataInputs.jsx';
+import AnonymousOrChannelSelect from './AnonymousOrChannelSelect.jsx';
+import { connect } from 'react-redux';
+import { getCookie } from '../utils/cookies.js';
+import {selectFile, clearFile, updateLoggedInChannel, updatePublishStatus, updateError} from '../actions';
+import PropTypes from 'prop-types';
+
+const LOAD_START = 'LOAD_START';
```
kauffj: These consts need to be shared across files

react/utils/page.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  replaceChannelSelectionInNavBar (loggedInChannel) {
```
kauffj: Shouldn't this design be obviated by addition of React/Redux? The component should be able to refresh based on external state change

react/utils/file.js
```diff
@@ -0,0 +1,38 @@
...
+      case 'image/png':
+        if (file.size > 10000000) {
+          console.log('file was too big');
+          throw new Error('Sorry, images are limited to 10 megabytes.');
+        }
+        break;
+      case 'image/gif':
+        if (file.size > 50000000) {
+          console.log('file was too big');
+          throw new Error('Sorry, .gifs are limited to 50 megabytes.');
```
kauffj: GIFs

react/reducers/index.js
```diff
@@ -0,0 +1,83 @@
+import {
+  CHANNEL_UPDATE, CLAIM_UPDATE, ERROR_UPDATE, FILE_CLEAR, FILE_SELECTED, METADATA_UPDATE, PUBLISH_STATUS_UPDATE,
+  SET_PUBLISH_IN_CHANNEL,
```
kauffj: `import * as`

react/reducers/index.js
```diff
@@ -0,0 +1,83 @@
...
+    name   : null,
+    shortId: null,
+    longId : null,
+  },
+  publishInChannel: false,
+  status          : {
+    status : null,
+    message: null,
+  },
+  error: {
```
kauffj: Some of these should possibly be renamed or refactored into separate files (or possibly can wait until next refactor/revision).

react/containers/PublishUrlInput.jsx
```diff
@@ -0,0 +1,117 @@
+import React from 'react';
+import { updateClaim } from '../actions/index';
+import { connect } from 'react-redux';
+import { makeGetRequest } from '../utils/xhr.js';
+import UrlMiddle from '../components/PublishUrlMiddle.jsx';
```
kauffj: You should look at some of the linting additions @IGassman has added to app, could help keep these imports consistent in path specification (among other improvements).

react/containers/PublishThumbnailInput.jsx
```diff
@@ -0,0 +1,92 @@
...
+        reject('error');
+      };
+      img.onload = function () {
+        clearTimeout(timer);
+        resolve('success');
+      };
+      timer = setTimeout(function () {
+        // reset .src to invalid URL so it stops previous
+        // loading, but doesn't trigger new load
+        img.src = '//!!!!/test.jpg';
```
kauffj: I suspect there is a less hacky way to do this ;)

react/containers/PublishForm.jsx
```diff
@@ -0,0 +1,254 @@
...
+  license              : PropTypes.string.isRequired,
+  nsfw                 : PropTypes.bool.isRequired,
+  loggedInChannel      : PropTypes.object.isRequired,
+  publishInChannel     : PropTypes.bool.isRequired,
+  fileError            : PropTypes.string,
+  urlError             : PropTypes.string,
+  publishSubmitError   : PropTypes.string,
+  onFileSelect         : PropTypes.func.isRequired,
+  onFileClear          : PropTypes.func.isRequired,
+  onChannelLogin       : PropTypes.func.isRequired,
```
kauffj: I suspect this shouldn't be necessary with addition of Redux

react/containers/AnonymousOrChannelSelect.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import { setPublishInChannel } from '../actions/index';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
+class AnonymousOrChannelSelect extends React.Component {
```
kauffj: I think just `ChannelSelect` is a fine name for this. "None" is just a choice.

react/components/ProgressBar.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+
+function ActiveBar () {
+  return <span className="progress-bar progress-bar--active">| </span>;
```
kauffj: What is the `|` doing here?

views/index.handlebars
```diff
@@ -1,52 +1,9 @@
...
-        </div>
-    </div>
-    <div id="publish-status" class="hidden">
-        <div class="row row--margined">
-            <div id="publish-update" class="row align-content-center"></div>
-            <div id="publish-progress-bar" class="row align-content-center"></div>
-            <div id="upload-percent" class="row align-content-center"></div>
+    <div id="react-publish-tool" class="row row--padded row--tall flex-container--column">
+        <div class="row row--padded row--tall flex-container--column flex-container--center-center">
+            <img src="/assets/img/loading.gif" alt="loading">
```
kauffj: This doesn't seem exactly on theme. Maybe ask nizuka?

react/components/ProgressBar.jsx
etisdew: It looks as if its seeding the progress visually, is that correct?

react/utils/file.js
etisdew: I'm interested in the direction of this PR, it seems to be purposed as an image sharing service as per required. .png .jpeg .gif ... requirements. Can we still publish video and is there support for generic file types? It just feels like its moving away from video upload and I wanted to be clear.

react/components/Preview.jsx
```diff
@@ -0,0 +1,57 @@
...
+    super(props);
+    this.state = {
+      imgSource       : '',
+      defaultThumbnail: '/assets/img/video_thumb_default.png',
+    };
+    this.previewFile = this.previewFile.bind(this);
+  }
+  componentWillMount () {
+    console.log('Preview will mount');
+    this.previewFile(this.props.file);
```
seanyesmunt: Generally you should avoid doing dom stuff/setting state in `componentWillMount`. Especially since it seems that they will be removing it in a future React version.

`componentDidMount` is preferred

react/components/ProgressBar.jsx
```diff
@@ -0,0 +1,76 @@
...
+      incrementer: 1,
+    };
+    this.startProgressBar = this.startProgressBar.bind(this);
+    this.updateProgressBar = this.updateProgressBar.bind(this);
+    this.stopProgressBar = this.stopProgressBar.bind(this);
+  }
+  componentWillMount () {
+    const bars = [];
+    for (let i = 0; i <= this.state.size; i++) {
+      bars.push(<InactiveBar key={i} />);
```
seanyesmunt: Instead of adding react elements to state, just store the data.

Then in the `render` function, you can do
`this.state.bars.map((bar, index) => bar.isActive ? <ActiveBar key={index} /> : <InactiveBar key={index} />)`

react/containers/ChannelLoginForm/view.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import { makePostRequest } from 'utils/xhr';
+import { setUserCookies } from 'utils/cookies';
+import { replaceChannelSelectionInNavBar } from 'utils/page';
+
+class ChannelLoginForm extends React.Component {
```
seanyesmunt: Forms are one of the most painful things in React. It might be worth looking into 3rd party React form elements. I've done a fair bit of work with forms and think that Formik is really nice. Hoping to add it to more parts of the app soon. https://github.com/jaredpalmer/formik

I really like how they use the `render` prop, I think it's a really nice way to interact with React elements

react/containers/ChannelSelect/view.jsx
```diff
@@ -0,0 +1,78 @@
...
+  constructor (props) {
+    super(props);
+    this.state = {
+      selectedOption: LOGIN,
+    };
+    this.toggleAnonymousPublish = this.toggleAnonymousPublish.bind(this);
+    this.handleSelection = this.handleSelection.bind(this);
+    this.selectOption = this.selectOption.bind(this);
+  }
+  componentWillMount () {
```
seanyesmunt: Same here, this should be moved to `componentDidMount`

react/containers/ChannelSelect/view.jsx
```diff
@@ -0,0 +1,78 @@
...
+    console.log('ChannelSelector will mount');
+    if (this.props.loggedInChannelName) {
+      this.selectOption(this.props.loggedInChannelName);
+    }
+  }
+  componentWillReceiveProps ({ loggedInChannelName }) {
+    if (loggedInChannelName) {
+      this.selectOption(loggedInChannelName);
```
seanyesmunt: Instead of reading prop values and adding them to state, you should just use the prop values and not worry about internal state.

I try to avoid internal state whenever possible, one exception is the app homepage, where we store if the user can scroll left or right in a card row. But that isn't dependent on any prop value, just by what is currently on the screen.

Whenever you separate the data into two sources it can cause some weird issues.

seanyesmunt:
Left a few comments. Looking good!

react/utils/file.js
billbitt: @etisdew yes `video/mp4` is supported and supporting video uploads is a key goal for spee.ch

react/utils/page.js
billbitt: The reason for this was that html outside of the react publish component (the top nav bar) needed to be updated.  To solve this, I added a nav bar component that can draw from the store.

react/components/ProgressBar.jsx
billbitt: Yes, I am rendering the `|`s for the progress bars

react/containers/ChannelSelect/view.jsx
billbitt: The issue I am struggling with, is that I (think) I need the `<select>` element to be a controlled component, and if so then it seems like it would be best to use internal state rather than the store, since no other component should care what option is selected unless this component decides to update the state of the store.  The reason I am setting the state when the component mounts and when it receives props, is that if another component updates the store when a new channel is logged into, then I want this `<select>` element to re-render with that option selected.  Any ideas on how I can approach this differently so that I can eliminate internal state while still accomplishing those goals?

react/components/NavBarChannelDropdown.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+function Logo ({ channelName, handleSelection, VIEW, LOGOUT }) {
```
kauffj: Misnamed

react/utils/xhr.js
```diff
@@ -0,0 +1,41 @@
...
+  makeGetRequest (url) {
+    return new Promise((resolve, reject) => {
+      let xhttp = new XMLHttpRequest();
+      xhttp.open('GET', url, true);
+      xhttp.responseType = 'json';
+      xhttp.onreadystatechange = () => {
+        if (xhttp.readyState === 4) {
+          if (xhttp.status === 200) {
+            resolve(xhttp.response);
+          } if (xhttp.status === 401) {
```
kauffj: All get and post requests should not assume a 401 means invalid username and password. Instead, have the server pass back the error message. Also, checks for 401 vs. 403 appear to be inconsistent.

public/assets/js/generalFunctions.js
```diff
@@ -40,25 +40,23 @@ function postRequest (url, params) {
...
-	}
-}
+    event.preventDefault();
+    var dataSet = event.target.dataset;
+    var status = dataSet.open;
+    var masterElement = document.getElementById(event.target.id||event.srcElement.id);
+    var slaveElement = document.getElementById(dataSet.slaveelementid);
+    var closedLabel = dataSet.closedlabel;
+    var openLabel = dataSet.openlabel;
+    if (status === "false") {
```
kauffj: This probably should not be a string based check.

public/assets/js/generalFunctions.js
```diff
@@ -131,16 +129,16 @@ function clearCookie(name) {
     document.cookie = `${name}=; expires=Thu, 01-Jan-1970 00:00:01 GMT;`;
 }
 
-function setUserCookies(channelName, channelClaimId, shortChannelId) {
-    setCookie('channel_name', channelName)
-    setCookie('channel_claim_id', channelClaimId);
-    setCookie('short_channel_id', shortChannelId);
+function setUserCookies(channelName, shortChannelId, channelClaimId) {
```
kauffj: We should look at getting all of this out of the cookie entirely.

Issue #324, opened 2018-01-18
Hardcoded LBRY API JSON address


maximest-pierre:
Inside helpers/lbryApi.js the API location is always Hardcoded. This means that the LBRY daemon must always be ```localhost:5279/lbryapi```. if we set it with a config file, we can split the dependency to multiple machines for deployment. This would also help to make a replicable dev environment with docker.

kauffj:
Good call @maximest-pierre! This should probably be in the `lbrySettings` part of `package.json`.

maximest-pierre:
I am probably going to fix this today. When I get a chance.

kauffj:
Oops, this isn't `lbry-app`. I suppose it would go in speechConfig.js? @billbitt should affirm this.

maximest-pierre:
I think this will go here ```spee.ch/config/speechConfig.js.example```

PR #325, opened 2018-01-18
Remove the hardcoded lbry API address

maximest-pierre:
This fixes issue #324.
The LBRY is no longer hardcoded to localhost and is set by the config file. 

I haven't tested it. I ran into some problems while trying to install spee.ch locally on my computer.

### DIFF
--- a/config/speechConfig.js.example
+++ b/config/speechConfig.js.example
@@ -35,4 +35,8 @@ module.exports = {
     testChannel        : '@testpublishchannel', // a channel to make test publishes in
     testChannelPassword: 'password',  // password for the test channel
   },
+  api: {
+    apiUri: 'localhost',
+    apiPort: '5279',
+  },
 };
--- a/helpers/lbryApi.js
+++ b/helpers/lbryApi.js
@@ -1,5 +1,8 @@
 const axios = require('axios');
 const logger = require('winston');
+const config = require('../config/speechConfig.js');
+const { apiUri, apiPort } = config.api;
+const lbryApiUrl = 'http://' + apiUri + ':' + apiPort;
 
 function handleLbrynetResponse ({ data }, resolve, reject) {
   logger.debug('lbry api data:', data);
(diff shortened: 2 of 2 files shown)


helpers/lbryApi.js
```diff
@@ -22,7 +24,7 @@ module.exports = {
     logger.debug(`lbryApi >> Publishing claim to "${publishParams.name}"`);
     return new Promise((resolve, reject) => {
       axios
-        .post('http://localhost:5279/lbryapi', {
+        .post('http://' + apiUri + ':' + apiPort + '/lbryapi', {
```
billbitt: Define the url string once at the top so the string doesn't have to be constructed repeatedly.  Also, the `/lbryapi` can be dropped.  
E.g. 
`const lbryUrl = 'http://' + apiUri + ':' + apiPort;`
...
`axios.post(lbryUrl, ... `

billbitt: Thanks for the PR!  

Define the url string once at the top so the string doesn't have to be constructed repeatedly.  Also, the `/lbryapi` can be dropped.  
E.g. 
`const lbryUrl = 'http://' + apiUri + ':' + apiPort;`
...
`axios.post(lbryUrl, ... `

If you can make this change I'll merge :)

billbitt: Also, let me know your discord handle so I can send you an lbc tip

maximest-pierre:
@billbitt I change it as requested.

config/speechConfig.js.example
```diff
@@ -35,4 +35,8 @@ module.exports = {
     testChannel        : '@testpublishchannel', // a channel to make test publishes in
     testChannelPassword: 'password',  // password for the test channel
   },
+  api: {
+    apiUri: 'localhost',
```
kauffj: If I'm being pedantic, which I am wont to do, this is more like a host than a URI. All of the parts together comprise the URI.

Here's a reference (this terminology is fairly universal, not google specific): https://www.mattcutts.com/blog/seo-glossary-url-definitions/

billbitt: I'm wont to do it correctly ;)

PR #329, opened 2018-01-23
Track publish time

billbitt:
- moved stats functions from `controllers` to `helpers` folder because they are more general utility functions that controllers
- added a function to send google analytics timing events
- added google timing event to publish api route


kauffj:
LGTM

PR #339, opened 2018-01-26
337 report link

billbitt:


kauffj:
@billbitt this shows merged but I don't seem to see it here? https://spee.ch/doineedablockchain

billbitt:
@kauffj I hadn't updated the spee.ch production server.  It is updated now.  There is a `Report` link under the claim details if the user clicks `[more]`.  Is that to inconspicuous?  I was having trouble coming up with where to place the link that would look good.  I guess the whole right hand column (description, links, social sharing buttons, etc.) could use a bit of design.

kauffj:
@billbitt that's probably fine but possibly some whitespace below the metadata.

PR #351, opened 2018-02-15
295 - server-side GA events for tracking serve requests

billbitt:
+ updated google analytics event and user timing actions
+ DRY'd up serve route code a little
+ added google analytics event actions to serve routes


billbitt:
note: a separate pull request will be made for adding the GA page view tracking to the react app

PR #360, opened 2018-02-23
350 open graph react

billbitt:
+ added server side rendering
+ added `react-helmet` to handle meta tags


webpack.config.js
```diff
@@ -1,32 +1,76 @@
 const Path = require('path');
```
seanyesmunt: I would recommend creating separate webpack configs for dev/prod. Before we did our big restructure changes the apps webpack setup had three files:

A base config: contains the generic stuff like babel, entry/output points, css loader, etc.
A dev config which extends the base config: contains stuff like `watch: true` (not needed in prod) and source maps
A prod config which extends the base config: contains stuff like minification (that's the main thing)

`webpack-merge` is a great tool for this
https://github.com/survivejs/webpack-merge

react/containers/LoginPage/view.jsx
```diff
@@ -13,26 +17,30 @@ class PublishPage extends React.Component {
     }
   }
   render () {
+    const pageTitle = createPageTitle('Login');
+    const canonicalLink = createBasicCanonicalLink('login');
+    const metaTags = createBasicMetaTags();
```
seanyesmunt: You can move all of these utility functions into the `SEO` component. Just pass in the minimum values needed and create the actual title/link/tags inside the component.

seanyesmunt: Then you don't need to add the same three lines to every page component.
`<SEO title="Login" link="login"  />` 
or
`<SEO title="Channel" link="channel"  channel={channel} />` 

Then in the component you can check if `channel` or `asset` was passed in and build the meta tags accordingly. 
```
if (!props.asset || !props.channel) {
    metaTags = createBasicMetaTags();
}
```

react/components/OpenGraphTags/index.jsx
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
+import View from './view';
+
+const mapStateToProps = ({ show }) => {
```
seanyesmunt: I'm probably missing something here, but wherever this is rendered, don't you already have the asset? You could just directly pass it in (or just pass some identifier to easily find it in state). Then you wouldn't need the `show` reducer here, which I don't really think needs to exist (in the future).

Then you wouldn't have to worry about the `show` reducer being populated and could use these tags for other pages (not sure if you would ever need to do that though)

models/index.js
```diff
@@ -30,16 +30,30 @@ sequelize
...
+// // add each model to the db object
+// fs
+//   .readdirSync(__dirname)
+//   .filter(file => {
+//     return (file.indexOf('.') !== 0 && file !== basename && file.slice(-3) === '.js');
+//   })
+//   .forEach(file => {
+//     const model = sequelize['import'](path.join(__dirname, file));
+//     db[model.name] = model;
+//   });
```
seanyesmunt: I think you can delete this?

react/components/OpenGraphTags/index.jsx
billbitt: ah, yes.  This component should actually be deleted.  I made it before moving everything into the `<SEO />` component.

PR #371, opened 2018-03-05
282 thumbnail selector

billbitt:
- for publishing videos: added a slider to select a thumbnail from the video
- moved the publish functionality out of the publish component and into a Redux saga
- added check on the back end publish route to check for and publish the thumbnail file if present


react/actions/publish.js
```diff
@@ -71,3 +71,18 @@ export function toggleMetadataInputs (showMetadataInputs) {
     data: showMetadataInputs,
   };
 };
+
+export function onNewThumbnail (file) {
+  console.log('new thumbnail action created');
```
seanyesmunt: Not needed

react/channels/publish.js
```diff
@@ -0,0 +1,52 @@
+import {buffers, END, eventChannel} from 'redux-saga';
+
+export const makePublishRequestChannel = (fd) => {
+  console.log('making publish request');
```
seanyesmunt: Not needed

routes/api-routes.js
```diff
@@ -135,14 +135,14 @@ module.exports = (app) => {
     logger.debug('api/claim-publish body:', body);
     logger.debug('api/claim-publish files:', files);
     // define variables
-    let  name, fileName, filePath, fileType, nsfw, license, title, description, thumbnail, channelName, channelPassword;
+    let  name, fileName, filePath, fileType, thumbnailFileName, thumbnailFilePath, thumbnailFileType, nsfw, license, title, description, thumbnail, channelName, channelPassword;
```
seanyesmunt: Why does spee.ch need to know the `thumbnailFilePath` on the server?

billbitt: This is because I'm having the client send the the thumbnail file along with the main video file together from the front end and creating the thumbnail data dynamically (if needed) based off of the main video file's metadata on the back end.

billbitt:
- removed the unneeded `browserHistory.js` file 
- cleaned up the console logs
- the vendor files were left over from when I was using masonry.js; I may add this back but they are not being used currently so I removed them

PR #374, opened 2018-03-06
refactor(google analytics): split ga timing events into multiple categories

billbitt:
in order to sharpen the google analytics surrounding events, I broke the events into two categories:
1. 'end-to-end' events which measure the time from receiving a request at the api endpoint to the time a response is issued to the client. (currently covers the publish api endpoint, with labels by file type)
2. 'lbrynet' events which measure the time from when a request is made to the daemon the time when the response is received from the daemon. (expanded these events with a label for each daemon action)

### DIFF
--- a/helpers/googleAnalytics.js
+++ b/helpers/googleAnalytics.js
@@ -13,11 +13,11 @@ function createServeEventParams (headers, ip, originalUrl) {
 };
 
 function createPublishTimingEventParams (category, variable, label, startTime, endTime) {
-  const durration = endTime - startTime;
+  const duration = endTime - startTime;
   return {
     userTimingCategory    : category,
     userTimingVariableName: variable,
-    userTimingTime        : durration,
+    userTimingTime        : duration,
     userTimingLabel       : label,
   };
 };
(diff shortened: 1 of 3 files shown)


helpers/googleAnalytics.js
```diff
@@ -13,31 +12,28 @@ function createServeEventParams (headers, ip, originalUrl) {
   };
 };
 
-function createPublishTimingEventParams (label, startTime, endTime, ip, headers) {
+function createPublishTimingEventParams (category, variable, label, startTime, endTime) {
   const durration = endTime - startTime;
```
kauffj: extra "r"

helpers/googleAnalytics.js
```diff
@@ -51,12 +47,14 @@ module.exports = {
...
-    sendGoogleAnalyticsTiming(ip, params);
+  sendGATimingEvent (category, variable, label, startTime, endTime) {
+    const params = createPublishTimingEventParams(category, variable, label, startTime, endTime);
+    sendGoogleAnalyticsTiming(siteName, params);
   },
-  sendGAChannelPublishTiming (headers, ip, originalUrl, startTime, endTime) {
-    const params = createPublishTimingEventParams('PUBLISH_IN_CHANNEL_CLAIM', startTime, endTime, ip, headers);
-    sendGoogleAnalyticsTiming(ip, params);
+  chooseGaLbrynetPublishLabel (publishParams) {
+    if (publishParams.channel_name || publishParams.channel_id) {
```
kauffj: could be a ternary

Issue #375, opened 2018-03-06
Improve dev ops and automated testing


kauffj:
@Igassmann has been adding a bunch of great features to the app. It'd be good to look at how many of these would be useful for spee.ch as well. Ones to consider:

- Prettier (if spee.ch doesn't have it)
- Linting
- Codacy
- DANGERFILE / CHANGELOG

billbitt:
I converted this to an epic so each piece could be an issue (prettier, linting, codacy, etc.)

skhameneh:
DANGERFILE is a Ruby CI thing, closing and leaving the Codacy task open for discussion

PR #382, opened 2018-03-08
fix: updated claim name availability check to support multiple claim addresses

billbitt:
- updated `claimNameIsAvailable` to check the Claim table rather than File table
- updated  `claimNameIsAvailable`  to be able to check with multiple claim addresses
- added a google analytics timing event for the `api/claim/availability` and `api/channel/availability` endpoints


react/containers/ChannelCreateForm/view.jsx
```diff
@@ -107,7 +90,7 @@ class ChannelCreateForm extends React.Component {
         this.props.onChannelLogin(result.channelName, result.shortChannelId, result.channelClaimId);
       })
       .catch((error) => {
-        this.setState({'error': error.message, status: null});
+        if (error.message) this.setState({'error': error.message, status: null});
```
kauffj: Always use brackets for if statements IMO.

Also - is Prettier and/or a linter not set up for this project? It can catch and even fix ones like this.

controllers/publishController.js
```diff
@@ -88,36 +88,43 @@ module.exports = {
   },
   claimNameIsAvailable (name) {
     // find any records where the name is used
-    return db.File.findAll({ where: { name } })
```
kauffj: I'm confused by this function, though I would have been equally confused before this changeset. If we're querying our own DB, why do we have to consider just the ones with a certain address? Is that because the DB contains all claims?

Also, if the addresses are in the DB, can't this filtering be done in SQL rather than JS?

react/containers/ChannelCreateForm/view.jsx
billbitt: yes to linter, not yet to Prettier (#375)

controllers/publishController.js
billbitt: Yes, the table that is being queried contains all claims, so we have to filter out claims belonging to this address.  I wonder if there would be reason to create a separate table for claims the site owns?  We avoided doing that previously because it seemed to be unnecessarily duplicative.

PR #383, opened 2018-03-08
fix: updated components to use single quotes in jsx

billbitt:
Per discussion in slack, we needed to decide on single or double quotes for html elements in the jsx.  I like single quotes because it reminds me that it is jsx, and thus other jsx rules need to be followed (e.g. using `className` rather than `class`).  This PR finishes converting the components to use single quotes.  I also moved all connected components into the `containers` folder.


react/components/ExpandingTextArea/index.jsx
```diff
@@ -7,12 +7,12 @@ class ExpandingTextarea extends Component {
...
   render () {
-    const { onChange, ...rest } = this.props;
+    const { ...rest } = this.props;
     return (
       <textarea
-        { ...rest }
+        {...rest}
         ref={x => this.el = x}
-        onChange={ this._handleChange.bind(this) }
+        onChange={this._handleChange.bind(this)}
```
seanyesmunt: You should bind handlers in the component constructor instead of inside the render function

billbitt: The `xxx/index` was an accident; my IDE attempts to update paths when I move files between folders.  I fixed the handler binding as well.  Thanks!

PR #385, opened 2018-03-10
[FIX #356] added html5 file format ( .ogg and .webm )

### LINKED ISSUE #356: Add HTML5 video file format to spee.ch
Add webm and ogg. mp4 is currently implemented.

ykris45:
maybe it can help issues and correct


billbitt:
@ykris45 thanks for you work on this. I am going to close this for now, as some of the code is going to be restructured, and then we can revisit this when we are ready to put it in the sprint.

Issue #377, opened 2018-03-07
Add LBRY app download link to spee.ch content pages


finer9:
near where it says 'hosted by spee.ch'

billbitt:
I'm actually not a huge fan of putting this on the asset pages, because I think it is good to keep a separation between spee.ch - and any other lbry implementations - and the app.  I think having a link to download on the `spee.ch/about` page is sufficient.  lmk if you disagree @finer9.

jsigwart:
I disagree - right now spee.ch is an excellent functional demo of a LBRY implementation. There is no confusion between spee.ch branding and LBRY - they are very distinct. Not promoting the LBRY technology on every image page is a missed opportunity.

billbitt:
ok, I'm convinced @jsigwart ;)

PR #387, opened 2018-03-12
add download link

ykris45:
Add LBRY app download link to spee.ch content pages #377


billbitt:
@finer9 this pr from @ykris45 would add a link to the about page.  Is that what you were thinking with #377?  My understanding of your issue is that you were requesting a link on the asset page itself.  Can you expand a little more on where and what we should link?

kauffj:
@billbitt this is fine to merge, but the original ticket was to add it to the show page

PR #388, opened 2018-03-12
Publish disable config

billbitt:
added a config variable with which publishing can be disabled


kauffj:
@billbitt this is an example of something we can no longer do in the model of spee.ch as a codebase to be used by others.

Instead, we'd need to do a design where spee.ch includes a component called `<PublishDisabledMessage>` (or whatever). This component contains a universal message like: "Publishing is disabled". It presumably shouldn't state that it's temporary. 

Then www.spee.ch would include a component with this text that overrides the default one.

PR #393, opened 2018-03-13
Create what-is-speech.md

LavRadis:
Just a start of spee.ch FAQ


kauffj:
@LavRadis for certain definitions of "start" ;)

PR #401, opened 2018-03-22
Speech as a package - export components

billbitt:
- added the Components, Containers, and Pages to the exported Speech class


billbitt:
@seanyesmunt  this PR might look a little strange to you because I have been working on this 'speech-as-a-package' branch that exports spee.ch as a constructor.  The purpose of this PR was to export all of the components (which are in the client/components, client/containers, and client/pages folders) along with the server so that they can be used to build custom components in spee.ch implementations.

build/getFolderNames.js
```diff
@@ -0,0 +1,10 @@
+const { lstatSync, readdirSync } = require('fs');
+const { join } = require('path');
+
+export const getSubDirectoryNames = (root) => {
+  return readdirSync(root)
+    .filter(name => {
+      let fullPath = join(root, name);
```
seanyesmunt: I would use `const` if you aren't changing anything

billbitt: nice

Issue #389, opened 2018-03-12
White lines on top and bottom in full screen mode 


LavRadis:
While watching https://spee.ch/e/heli.mp4 I've noticed a white line on the top and bottom of the screen. Tom sees only one on top, Chris sees both. I'm using Chrome on Win 10. 
![untitled](https://user-images.githubusercontent.com/10566738/37309929-bab46538-2642-11e8-9de5-33bf01811449.jpg)

billbitt:
I've also noticed that on videos with these proportions, the 'hosted via spee.ch' link gets covered by the video

billbitt:
@LavRadis the white line issue you are describing is when the video is full-screened, correct?

LavRadis:
@billbitt  yes.

billbitt:
@LavRadis oops, I'm an idiot, you said 'full screen' in the title of the issue

Smurphonomics:
Fixed here: #414

PR #414, opened 2018-04-17
Fullscreen Border & Padding Fix

Invariant-Change:
Fixes border AND padding which are displaying in fullscreen mode. 

See: #389


billbitt:
@Invariant-Change we're splitting the code base so that this repo will house the server code, while the app and components will be in separate repos.  Currently the css is in https://github.com/lbryio/www.spee.ch; if you could make this pr on that repo, that would be great.

PR #416, opened 2018-04-17
Speech as a package

billbitt:
This pr converts the codebase to a package which exports a customizeable `Server` class that can be consumed and used by `www.spee.ch` and other spee.ch implementations


README.md
```diff
@@ -1,54 +1,19 @@
...
-  * `server.js` is the entry point for the server.  It creates the [express app](https://expressjs.com/), requires the routes, syncs the database, and starts the server listening on the `PORT` designated in the config file. This file is the entry point for webpack to build the server bundle.
-  * the `/routes` folder contains all of the routes for the express app
-  * the `/models` folder contains all of the models which the app uses to interact with the `mysql` database.  Note: this app uses the [sequelize](http://docs.sequelizejs.com/) orm.
+## Development / Structure 
+* the `server/` folder contains all of the server code
+  * `index.js` is the entry point for the server.  It creates the [express app](https://expressjs.com/), requires the routes, syncs the database, and starts the server listening on the `PORT` designated in the config file. This file is the entry point for webpack to build the server bundle.
+  * the `server/routes` folder contains all of the routes for the express app
+  * the `server/models` folder contains all of the models which the app uses to interact with the `mysql` database.  Note: this app uses the [sequelize](http://docs.sequelizejs.com/) orm.
```
seanyesmunt: ORM

config/mysqlConfig.js
```diff
@@ -0,0 +1,20 @@
...
+    // configure credentials
+    logger.info('configuring mysql...');
+    const { database, username, password } = config;
+    this.database = database;
+    this.username = username;
+    this.password = password;
+  };
+}
+
+module.exports = new mysql();
```
seanyesmunt: This should probably be `new Mysql()` since it is a class function

index.js
```diff
@@ -0,0 +1,2 @@
...eturn()}finally{if(a)throw i}}return t}return function(n,t){if(Array.isArray(n))return n;if(Symbol.iterator in Object(n))return e(n,t);throw new TypeError("Invalid attempt to destructure non-iterable instance")}}(),a=t(4),i=a.sendGAServeEvent,o=t(23),l=o.determineResponseType,s=o.flipClaimNameAndIdForBackwardsCompatibility,u=o.logRequestData,c=o.getClaimIdAndServeAsset,d=t(24),f=t(25),h=function(e,n){var t=e.headers,a=e.ip,o=e.originalUrl,h=e.params,m=void 0;try{m=d.parseModifier(h.claim).hasFileExtension}catch(e){return n.status(400).json({success:!1,message:e.message})}var p=l(m,t);if("SERVE"!==p)return f(e,n);i(t,a,o);var g=void 0;try{g=d.parseClaim(h.claim).claimName}catch(e){return n.status(400).json({success:!1,message:e.message})}var v=void 0,y=void 0,b=void 0,w=void 0;try{var C=d.parseIdentifier(h.identifier);v=C.isChannel,y=C.channelName,b=C.channelClaimId,w=C.claimId}catch(e){return n.status(400).json({success:!1,message:e.message})}if(!v){var I=s(w,g),x=r(I,2);w=x[0],g=x[1]}u(p,g,y,w),c(y,b,g,w,o,a,n)};e.exports=h},function(e,n,t){"use strict";var r=t(90);e.exports=function(e){e.get("*",r)}},function(e,n,t){"use strict";var r=t(14),a=function(e,n){r(e,n)};e.exports=a}]);
```
seanyesmunt: Don't commit this

index.js.map
```diff
@@ -0,0 +1 @@
...S/C,EAAQ,EAEvBH,GAAOC,SACLqoB,iBADe,SACG1d,EAAaE,EAAW6d,EAAiB1I,GAEzD,IAAKrV,IAAgBE,EACnB,OACEF,YAAgB,KAChB0C,eAAgB,KAIpB,IAAI2S,EAAM,CACR,GAAIrV,GAAeA,IAAgBqV,EAAKrV,YACtC,KAAM,IAAIgB,OAAM,4DAElB,IAAId,GAAaA,IAAcmV,EAAK3S,eAClC,KAAM,IAAI1B,OAAM,0DAElB,QACEhB,YAAgBqV,EAAKrV,YACrB0C,eAAgB2S,EAAK3S,gBAIzB,IAAKqb,EAAiB,KAAM,IAAI/c,OAAM,+BACtC,OAAO5L,GAAOC,QAAQgpB,+BAA+Bre,EAAaE,EAAW6d,IAE/EM,+BA1Be,SA0BiBre,EAAaE,EAAWoe,GACtD,MAAO,IAAI9c,SAAQ,SAACZ,EAASC,GAE3B,GAAIkZ,UAEAwE,IACAve,KAAaue,EAAA,YAAmCve,GAChDE,IAAWqe,EAAA,eAAsCre,GAErD7E,EAAGzB,QACAkC,SACCC,MAAOwiB,IAERrjB,KAAK,SAAAqK,GACJ,IAAKA,EAEH,KADAjN,GAAO2D,MAAM,oBACP,GAAI+E,OAAM,gEAIlB,OAFA+Y,GAAcxU,EAAQhP,MACtB+B,EAAO2D,MAAM,gBAAiB8d,GACvB1e,EAAGrB,KAAK8B,SACbC,OAASiZ,SAAU+E,EAAY/Z,YAAYkJ,UAAU,QAGxDhO,KAAK,SAAAma,GACJ,IAAKA,EAEH,KADA/c,GAAO2D,MAAM,iBACP,GAAI+E,OAAM,gEAElB,OAAOqU,GAAKC,gBAAgBgJ,KAE7BpjB,KAAK,SAAAqa,GACJ,IAAKA,EAEH,KADAjd,GAAO2D,MAAM,sBACP,GAAI+E,OAAM,gEAElB1I,GAAO2D,MAAM,8BACb2E,EAAQmZ,KAET3e,MAAM,SAAA1C,GACLmI,EAAOnI,U1EsmHX,SAAUtD,EAAQC,EAASE,GAEjC,YAGA,IAAI2E,G2EhrHmB3E,EAAQ,GAAvB0M,E3EirHS/H,E2EjrHT+H,W3EmrHJvB,E2ElrH4BnL,EAAQ,GAAhCgD,E3EmrHkBmI,E2EnrHlBnI,oBAQFgjB,EAAe,SAAAxb,EAAuCpH,GAAQ,GAAnCF,IAAmCsH,EAA5
```
seanyesmunt: This one too

server/index.js
```diff
@@ -0,0 +1,115 @@
...
+const logger = require('winston');
+const requestLogger = require('middleware/requestLogger.js');
+const Path = require('path');
+const loggerConfig = require('loggerConfig.js');
+const mysqlConfig = require('mysqlConfig.js');
+const siteConfig = require('siteConfig.js');
+const slackConfig = require('slackConfig.js');
+
+function Server () {
+  this.configureLogger = (userConfig) => {
```
seanyesmunt: For these you can just do 
```
this.configureLogger = loggerConfig.update
```

server/index.js
```diff
@@ -0,0 +1,115 @@
...
+    app.use(requestLogger);
+
+    // configure passport
+    const speechPassport = require('speechPassport');
+    // initialize passport
+    const sessionKey = siteConfig.auth.sessionKey;
+    app.use(cookieSession({
+      name  : 'session',
+      keys  : [sessionKey],
+      maxAge: 24 * 60 * 60 * 1000, // i.e. 24 hours
```
seanyesmunt: Do we need to have any `maxAge`? It would be nice if the user was never logged out

server/index.js
```diff
@@ -0,0 +1,115 @@
...
+    }));
+    app.use(speechPassport.initialize());
+    app.use(speechPassport.session());
+
+    // configure handlebars & register it with express app
+    const hbs = expressHandlebars.create({
```
seanyesmunt: I thought handlebars was removed?

index.js
billbitt: hmm, the problem is that the server uses jsx for server side rendering, so it needs to be transpiled (I think that's the only part that needs to be transpiled), so I wanted to build a bundle and export that bundle.  Do you think I should just transpile with babel and not actual bundle with webpack?  That's what I did for [spee.ch-components](https://github.com/lbryio/spee.ch-components)

seanyesmunt: Oh I gotcha. I was confused.

I still don't think this needs to be committed. You can build before publishing. IMO if someone is cloning the repo, they will be building the app on their own.

billbitt: i updated it to only transpile the two jsx render functions that are needed for SSR, the rest does not need to be built at this level, as the implementation will build the app again.

server/index.js
billbitt: it's still being used in one specific case: for sharing videos on twitter via cards, twitter needs an html page with a video player embedded, so I am using handlebars to generate that one page when needed.

server/index.js
billbitt: I guess we don't need to, and it does make a better user experience if the user doesn't need to log back in ever day.  I eliminated the maxAge so sessions should last indefinitely.

PR #434, opened 2018-05-04
Claim availability check

billbitt:
- updates to api routes to properly serve front end changes in this pr: https://github.com/lbryio/spee.ch-components/pull/4


billbitt:
yeah, I think there is some refactoring that can be done

PR #449, opened 2018-05-18
Startup fixes

billbitt:
- high level error detection checks for lbrynet and retrieves wallet balance on startup
- removed confusing test of error log
- added a step to create the database if it was not already created

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,6 +14,7 @@ const mysqlConfig = require('./config/mysqlConfig.js');
 const siteConfig = require('./config/siteConfig.js');
 const slackConfig = require('./config/slackConfig.js');
 const createDatabaseIfNotExists = require('./server/models/utils/createDatabaseIfNotExists.js');
+const { getWalletBalance } = require('./server/lbrynet');
 
 function Server () {
   this.configureLogger = loggerConfig.update;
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -81,15 +83,27 @@ function Server () {
     const db = require('./server/models');
     const PORT = siteConfig.details.port;
     // sync sequelize
-    db.sequelize.sync()
-    // start the server
+    createDatabaseIfNotExists()
+      getWalletBalance()
```
seanyesmunt: This is formatted weird

index.js
```diff
@@ -81,15 +83,27 @@ function Server () {
...
+      .then(balance => {
+        logger.info('starting LBC balance:', balance);
+      })
+      .then(() => {
+        db.sequelize.sync()
+      })
```
seanyesmunt: These can all live inside one `.then` after the `getWalletBalance()`

No need to break them up

billbitt: I tried updating a bit, is this what you had in mind?  I'm still a little fuzzy on the best patterns for chaining and when to use `.then`, but I suppose `db.sequelize.sync` and `this.server.listen` can be handled asynchronously

PR #464, opened 2018-05-31
Password reset

billbitt:
- added api route to allow for password reset via old password or master password in config

### DIFF
--- a/server/controllers/api/user/password/index.js
+++ b/server/controllers/api/user/password/index.js
@@ -12,7 +12,7 @@ const { auth: { masterPassword } } = require('../../../../../config/siteConfig.j
 const updateUserPassword = ({ ip, originalUrl, body }, res) => {
   let userRecord;
   const { userName, oldPassword, newPassword } = body;
-  logger.info('body:', body);
+
   if (!masterPassword) {
     return res.status(400).json({
       success: false,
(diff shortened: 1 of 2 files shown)


server/controllers/api/user/password/index.js
```diff
@@ -0,0 +1,69 @@
...
+    });
+  }
+
+  db.User.findOne({
+    where: {
+      userName,
+    },
+  })
+  .then(user => {
+    userRecord = user;
```
seanyesmunt: Why don't you just use `user` here?

server/controllers/api/user/password/index.js
```diff
@@ -0,0 +1,69 @@
+const { handleErrorResponse } = require('../../../utils/errorHandlers.js');
+const logger = require('winston');
+const db = require('../../../../models');
+const { auth: { masterPassword } } = require('../../../../../config/siteConfig.js');
```
seanyesmunt: This seems like it could be a security vulnerability?

server/controllers/api/user/password/index.js
billbitt: I need the `user` object to be available in the next `.then` in the chain, so I hoisted it up to the `updateUserPassword`'s scope by defining `userRecord` there.  Is there a better pattern?  I need the `user` object before I can check for a password match, and have to check for a password match before operating on that user object to change the password.

seanyesmunt: Oh duh.

server/controllers/api/user/password/index.js
billbitt: As discussed in standup, the siteConfig will not be kept in version control (it is pulled from www.spee.ch's config which is gitignored).  I'm going to merge.

PR #470, opened 2018-06-06
Fix configs loading

billbitt:
updated so all configs are pulled directly where needed via module aliasing


client/src/components/GAListener/index.jsx
```diff
@@ -2,28 +2,34 @@ import React from 'react';
...
-        }
+if (siteConfig.analytics) {
+  ({ googleId } = siteConfig.analytics);
+}
 
-        render () {
-            return this.props.children;
-        }
-    }
+GoogleAnalytics.initialize(googleId);
```
seanyesmunt: Should this be inside the `if` statement?

billbitt: Yes, good catch.  I also added logic to skip the sending of the analytics if no googleId present.

PR #476, opened 2018-06-14
vertically align channel contents and reverse order

billbitt:


client/src/utils/createGroupedList.js
```diff
@@ -0,0 +1,19 @@
+export const createGroupedList = (list, size) => {
+  if (!size) {
+    console.log('no size provided to createGroupedList');
```
seanyesmunt: Not needed.

client/src/utils/createGroupedList.js
```diff
@@ -0,0 +1,19 @@
...
+    return list;
+  }
+  if (!list) {
+    console.log('no list provided to createGroupedList');
```
seanyesmunt: Same here

billbitt: my concern was I didn't want the function to return early silently.  I replaced the above console logs with errors.

seanyesmunt:
Cool. Just squash your commits and it's good to go.

PR #484, opened 2018-06-18
Adding Border Back to Videos

Invariant-Change:
Video border was lost in recent [PR](https://github.com/lbryio/spee.ch/pull/471) This adds it back.


billbitt:
thanks @Invariant-Change !

PR #492, opened 2018-06-26
465 social share with extensions

billbitt:
adds open graph support for image urls with the file type suffix. e.g. spee.ch/a/example.jpg


server/controllers/assets/utils/determineRequestType.js
```diff
@@ -0,0 +1,61 @@
+const logger = require('winston');
+const { EMBED, BROWSER, SOCIAL } = require('../constants/request_types.js');
+
+function headersMatchesSocialBotList (headers) {
```
seanyesmunt: This could be simplified by using an object instead of an array
```
const socialBotList = {
    'facebookexternalhit': 1,
    'Twitterbot': 1,
}
​
if (socialBotList[userAgent]) {
  logger.debug();
  return true;
} else {
  return false;
}
```

billbitt: cool pattern.  I updated to use this.

billbitt:
@seanyesmunt previous URLs should not be negatively affected.  This adds a new route, but does not remove the existing routes.  The new route will be used only for the image sources for SEO tags, so it is specifically for serving images for SEO.

server/controllers/assets/utils/determineRequestType.js
```diff
@@ -0,0 +1,61 @@
...
+
+const determineRequestType = (hasFileExtension, headers) => {
+  let responseType;
+  logger.debug('headers:', headers);
+  // return early with 'show' if headers match the list
+  if (headersMatchesSocialBotList(headers)) {
+    return SOCIAL;
+  }
+  // if request is not from a social bot...
+  if (hasFileExtension) {
```
kauffj: Multiple return statements in a function can be a contentious topic (e.g. https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement), but I think a function like this is better if the returns are just inlined, rather than setting a variable.

server/controllers/assets/utils/determineRequestType.js
kauffj: @seanyesmunt do you prefer that to just:
```
const haystack = ['a', 'b']
if (haystack.includes(needle))
{
  ...
}
```

kauffj: (_technically_ `includes` is presumably `O(n)` and a map lookup `O(1)`, but this is such an efficient operation that this  doesn't matter)

PR #499, opened 2018-06-28
Tor middleware

billbitt:
- added a table to be used to store tor node ips
- added a middleware function which can be used to check request against tor node list


server/middleware/torCheckMiddleware.js
```diff
@@ -0,0 +1,24 @@
...
+  return db.Tor.findAll(
+    {
+      where: {
+        address: ip,
+      },
+      raw: true,
+    })
+    .then(result => {
+      logger.debug('tor check results:', result);
+      req['tor'] = (result.length >= 1); // add this to the req object
```
seanyesmunt: Maybe you should just stop the request here? Not sure, if there are other routes in that may be blocked in the future it will be simpler to add. If it's gonna be only the /publish route then this is probably fine.

PR #504, opened 2018-06-29
changed social bot list to array

billbitt:
changing social bot check back to an array instead of object, because the check needs to be non-exact.
in the below code, it only returns true if the user agent string is an exact match
```
const socialBotList = {
    'facebookexternalhit': 1,
    'Twitterbot': 1,
}
​
if (socialBotList[userAgent]) {
  logger.debug();
  return true;
} else {
  return false;
}
```
in the below version, it will catch any string that contains the social bot name.  e.g. ` facebookexternalhit/1.1 (+http://www.facebook.com/externalhit_u
atext.php)` or `Twitterbot/1.0`.  This will mean a change to `Twitterbot/2.0` or related will not require a change in this code.
```
const socialBotList = [
  'facebookexternalhit',
  'Twitterbot',
];
for (let i = 0; i < socialBotList.length; i++) {
  if (userAgent.indexOf(socialBotList[i]) >= 0) {
    logger.debug('request is from social bot:', socialBotList[i]);
    return true;
  }
}
```

### DIFF
--- a/server/controllers/assets/utils/determineRequestType.js
+++ b/server/controllers/assets/utils/determineRequestType.js
@@ -3,26 +3,32 @@ const { EMBED, BROWSER, SOCIAL } = require('../constants/request_types.js');
 
 function headersMatchesSocialBotList (headers) {
   const userAgent = headers['user-agent'];
-  const socialBotList = {
-    'facebookexternalhit': 1,
-    'Twitterbot'         : 1,
-  };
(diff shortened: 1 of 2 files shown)


seanyesmunt:
Ah that was my mistake. Guess I didn't look close enough. 

You could use ` if (userAgent.includes(socialBotList[i])) {` instead to make it a little simpler.

PR #506, opened 2018-07-02
Blocked timeout error

billbitt:
updated `blocked` table to use `outpoint` and keep all blocked claims

_note: requires dropping old blocked table prior to running this branch_

### DIFF
--- a/server/models/blocked.js
+++ b/server/models/blocked.js
@@ -6,14 +6,6 @@ module.exports = (sequelize, { STRING }) => {
   const Blocked = sequelize.define(
     'Blocked',
     {
-      claimId: {
-        type     : STRING,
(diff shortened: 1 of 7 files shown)


server/models/blocked.js
```diff
@@ -46,5 +37,42 @@ module.exports = (sequelize, { STRING }) => {
...
+        return this.destroy({
+          truncate: true,
+        });
+      })
+      .then(() => {
+        // fill the table
+        return this.bulkCreate(blockedList);
+      })
+      .catch(error => {
+        throw error;
```
seanyesmunt: Instead of throwing, catching, and throwing an error again, you can just skip this `catch` block and it will go to the next highest `catch`

PR #508, opened 2018-07-03
505 twitter share undefined

billbitt:
fixes bug where 'undefined' was showing up in social links rather than the claim name

### DIFF
--- a/client/src/components/ClickToCopy/index.jsx
+++ b/client/src/components/ClickToCopy/index.jsx
@@ -1,0 +1,37 @@
+import React from 'react';
+
+class ClickToCopy extends React.Component {
+  constructor (props) {
+    super(props);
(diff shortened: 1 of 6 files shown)


client/src/components/ClickToCopy/index.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+class ClickToCopy extends React.Component {
+  constructor (props) {
+    super(props);
+    this.copyToClipboard = this.copyToClipboard.bind(this);
+  }
+  copyToClipboard (event) {
+    const elementToCopy = event.target.id;
+    const element = document.getElementById(elementToCopy);
```
seanyesmunt: could this just be `const element = event.target;` ?

billbitt: I thought no, but the answer is yes :)

PR #513, opened 2018-07-07
fix app thumbnail viewing on spee.ch

billbitt:
This PR fixes the issue where spee.ch was showing a page with a broken link instead of the asset.  This issues seems to have been a result of cloudflare caching the page document and serving that instead of the image or video. 
- added check for file name in publish request
- fixed accuracy of some file check error messages
- remove default content type of `image/jpeg` on file serving
- changed embeded assets to use the `spee.ch/asset/<name>/<claimid>` route which is specific for serving the asset itself, thus avoiding the confusion which lead to cloudflare serving the document from the cache

### DIFF
--- a/client/build/components/AssetPreview/index.js
+++ b/client/build/components/AssetPreview/index.js
@@ -19,7 +19,7 @@ var AssetPreview = function AssetPreview(_ref) {
       fileExt = _ref$claimData.fileExt,
       contentType = _ref$claimData.contentType,
       thumbnail = _ref$claimData.thumbnail;
-  var directSourceLink = "".concat(claimId, "/").concat(name, ".").concat(fileExt);
+  var directSourceLink = "asset/".concat(name, "/").concat(claimId);
   var showUrlLink = "/".concat(claimId, "/").concat(name);
   return _react.default.createElement(_reactRouterDom.Link, {
     to: showUrlLink
--- a/client/build/containers/AssetDisplay/view.js
+++ b/client/build/containers/AssetDisplay/view.js
@@ -82,7 +82,7 @@ function (_React$Component) {
           case 'image/gif':
             return _react.default.createElement("img", {
               className: "asset-image",
-              src: "/".concat(claimId, "/").concat(name, ".").concat(fileExt),
+              src: "/asset/".concat(name, "/").concat(claimId),
               alt: name
             });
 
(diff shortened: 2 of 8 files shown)


client/build/components/AssetPreview/index.js
```diff
@@ -19,7 +19,7 @@ var AssetPreview = function AssetPreview(_ref) {
       fileExt = _ref$claimData.fileExt,
       contentType = _ref$claimData.contentType,
       thumbnail = _ref$claimData.thumbnail;
-  var directSourceLink = "".concat(claimId, "/").concat(name, ".").concat(fileExt);
+  var directSourceLink = "asset/".concat(name, "/").concat(claimId);
```
skhameneh: not a stopper, but it would be ideal to stick with single or double quotes, I see some mixed formatting

client/build/containers/AssetDisplay/view.js
```diff
@@ -82,7 +82,7 @@ function (_React$Component) {
           case 'image/gif':
             return _react.default.createElement("img", {
               className: "asset-image",
-              src: "/".concat(claimId, "/").concat(name, ".").concat(fileExt),
+              src: "/asset/".concat(name, "/").concat(claimId),
```
skhameneh: Using concat on strings is a little different that common convention, does it have a negative performance impact?
You could also use template literals

```
src: `/asset/${name}/${claimId}`

client/build/components/AssetPreview/index.js
kauffj: @billbitt I saw #405 was closed but if you add Prettier I believe this will never be an issue again.

billbitt: We should add Prettier (I reopened #405), but these files are the build result from babel transpiling the es6 code.
I'd like to avoid checking the built code into version control, but we have to include it at this point so that it can be imported by other packages like `lbryio/www.spee.ch`.  I'd like merge `lbryio/www.spee.ch` and `lbryio/spee.ch` (#516) which should result in being able to have one build process that does not need to check built code into version control.

client/build/containers/AssetDisplay/view.js
billbitt: see above, but this is the result of babel

PR #514, opened 2018-07-08
Error Message Fix

Invariant-Change:
## Error Message Fix

Fixes the error message issue shown in the below image.

### Before:
image

### After:
image

### DIFF
--- a/client/scss/components/_asset-display.scss
+++ b/client/scss/components/_asset-display.scss
@@ -1,8 +1,9 @@
 .asset-display {
   flex: 1 0 auto;
-  display: block;
+  display: flex;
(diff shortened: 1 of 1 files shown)


billbitt:
here's an example of the content no longer centered with `display` set to `block`.
screenshot 2018-07-08 20 47 36

Invariant-Change:
Another class would be good for better styling of the error message. However, I just updated the fix to keep alignment and fix the overflow while still using flex, which should now work without changing any other part of the site. Let me know if it's all fine your end :)

billbitt:
works great. Thanks @Invariant-Change !

PR #520, opened 2018-07-13
Fixes Missing Line & Stop Content Touchings Sides

Invariant-Change:
Fixes missing Line in the header and stops content touchings sides with the other fixes.

image

image
image


billbitt:
closing this @Invariant-Change because #517 appears to have fixed it.

PR #521, opened 2018-07-13
Margin Breaks Flex Vertical Alignment

Invariant-Change:
Removing the marking fixes vertical alignment issues when using Flex.


Invariant-Change:
I kept them in for backwards compatibility. Thought that maybe in the future, this could be replaced with a var.

PR #531, opened 2018-07-17
493 twitter video

billbitt:
patch for twitter video embed issue
- removes some dead handlebars code
- removes X-Frame-Option header which was blocking twitter video embed

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,6 @@ checkForConfig('siteConfig');
 const express = require('express');
 const bodyParser = require('body-parser');
 const expressHandlebars = require('express-handlebars');
-const Handlebars = require('handlebars');
 const helmet = require('helmet');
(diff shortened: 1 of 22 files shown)


index.js
```diff
@@ -81,11 +80,15 @@ function Server () {
     app.use(speechPassport.session());
 
     // configure handlebars & register it with express app
-    const hbs = expressHandlebars.create({
+    const viewsPath = Path.resolve(process.cwd(), 'node_modules/spee.ch/server/views');
+    app.engine('handlebars', expressHandlebars({
+      async: false,
```
skhameneh: what will it take to allow async?

billbitt: hmm, I'm actually not sure why it is set to false other than it was probably recommended in the `express-handlebars` implementation examples.  I'll look into that.

PR #532, opened 2018-07-17
updated serve and show urls in publish response

billbitt:
updating the publish api response to include both the `show` and `serve` urls for an asset


billbitt:
closing without merging because this change is better made via https://github.com/lbryio/spee.ch/pull/538

PR #538, opened 2018-07-18
serve static assets on urls with file extension

billbitt:
This PR updates the routes so that they break into two categories:
- any route without a file extension will be a `show` route, meaning that this route returns an html document with meta tags, etc.  These routes are 'dynamic' in that they can also serve just the asset (video or image) if the request headers indicate this. Example: `spee.ch/a/apple` or `spee.ch/apple`.
- any route with a file extension will be a `serve` route, meaning the route will return the asset (video, image, etc. ) file directly.  These routes are `static`, meaning they will return only the asset and not an html document regardless of the request headers. Examples: `spee.ch/a/apple.jpg` or `spee.ch/apple.jpg`.


tzarebczan:
What are the headers required to serve the static content on a url without a file extention?

billbitt:
@tzarebczan it assumes the asset is requested if the accept headers include `image/` or `video` and not `text/html`.  This could be added too or changed easily.  Like if we want to check for the app by looking for `LBRY/` in the `user-agent`.
```
const imageIsWanted = accept && accept.match(/image\/.*/) && !accept.match(/text\/html/);
const videoIsWanted = accept && accept.match(/video\/.*/) && !accept.match(/text\/html/);
```

client/build/utils/createCanonicalLink.js
```diff
@@ -0,0 +1,55 @@
...
+  if (asset.claimData) {
+    var _asset$claimData = asset.claimData;
+    channelName = _asset$claimData.channelName;
+    certificateId = _asset$claimData.certificateId;
+    name = _asset$claimData.name;
+    claimId = _asset$claimData.claimId;
+  }
+
+  if (channelName) {
+    return "".concat(host, "/").concat(channelName, ":").concat(certificateId, "/").concat(name);
```
skhameneh: use template strings?
e.g.
```
return `${host}/${channelName}:${certificateId}/${name}`
```

billbitt: This is the babel transpiled code :)

billbitt: once I merge the www.spee.ch and spee.ch repos we should be able to keep this out of version control

PR #539, opened 2018-07-19
added object-fit and position to assets

billbitt:
fix for images that are stretched too wide


client/scss/components/_asset-display.scss
```diff
@@ -9,6 +9,8 @@
 .asset-image, .asset-video {
   max-width: 100%;
   max-height: 100%;
+  object-fit: contain;
```
skhameneh: Order by alphabetical name? 😅

billbitt: :+1:

PR #546, opened 2018-07-30
add oEmbed support

billbitt:
- added `<link />` tags via Helmet which include oEmbed url
- added `api/oembed...` route for oEmbed requests


server/controllers/api/oEmbed/getOEmbedDataForAsset.js
```diff
@@ -0,0 +1,60 @@
...
+        where: {
+          name: claimName,
+          claimId,
+        },
+      });
+    })
+    .then(fileRecord => {
+      fileData = fileRecord.dataValues;
+      logger.debug('file data:', fileData);
+      const serveUrl = `https://dev1.spee.ch/${fileData.claimId}/${fileData.name}.${fileData.fileType.substring(fileData.fileType.indexOf('/') + 1)}`;
```
skhameneh: Do you have configs for the host URL generation?
That would be ideal over hardcoding

server/controllers/api/oEmbed/getOEmbedDataForChannel.js
```diff
@@ -0,0 +1,26 @@
...
+    .then(certificateRecord => {
+      const certificateData = certificateRecord.dataValues;
+      return {
+        version      : 1.0,
+        provider_name: 'Spee.ch',
+        provider_url : 'https://spee.ch',
+        type         : 'link',
+        author_name  : certificateData.name,
+        title        : `${certificateData.name}'s channel on Spee.ch`,
+        author_url   : `https://spee.ch/${certificateData.name}:${certificateData.claimId}`,
```
skhameneh: Same, use config instead of hard-coding

client/src/containers/SEO/view.jsx
```diff
@@ -17,17 +21,24 @@ class SEO extends React.Component {
...
         title={pageTitle}
         meta={metaTags}
-        link={[{rel: 'canonical', href: canonicalLink}]}
+        link={[
+          {
+            rel : 'canonical',
+            href: cannonicalLink,
+          },
+          oEmbed.json(host, cannonicalLink),
+          oEmbed.xml(host, cannonicalLink),
```
skhameneh: We should probably leave XML out until it's supported.

server/controllers/api/oEmbed/getOEmbedDataForAsset.js
billbitt: good catch.  I meant to do this after testing and forgot.

PR #552, opened 2018-08-02
498 incorrect short url

billbitt:
- added a method to the Claim model to check for max height
- updated Claim record creation on new publishes to use max height rather than 0


utils/lbryUri.js
```diff
@@ -56,15 +52,13 @@ module.exports = {
     };
   },
   parseClaim: function (claim) {
-    logger.debug('parsing name:', claim);
     const componentsRegex = new RegExp(
       '([^:$#/.]*)' + // name (stops at the first modifier)
       '([:$#.]?)([^/]*)' // modifier separator, modifier (stops at the first path separator or end)
     );
-    const [proto, claimName, modifierSeperator, modifier] = componentsRegex
+    const [, claimName, modifierSeperator, modifier] = componentsRegex
```
skhameneh: Remove the extra comma

utils/lbryUri.js
```diff
@@ -1,20 +1,16 @@
...
   REGEXP_ADDRESS        : /^b(?=[^0OIl]{32,33})[0-9A-Za-z]{32,33}$/,
   CHANNEL_CHAR          : '@',
   parseIdentifier       : function (identifier) {
-    logger.debug('parsing identifier:', identifier);
     const componentsRegex = new RegExp(
       '([^:$#/]*)' + // value (stops at the first separator or end)
       '([:$#]?)([^/]*)' // modifier separator, modifier (stops at the first path separator or end)
     );
-    const [proto, value, modifierSeperator, modifier] = componentsRegex
+    const [, value, modifierSeperator, modifier] = componentsRegex
```
skhameneh: Remove extra comma

utils/lbryUri.js
billbitt: I think it is necessary in order to ignore the first value when destructuring the returned array, right?  Otherwise claimName becomes the proto.. etc.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

PR #579, opened 2018-08-15
added node-ip and updated CLI

billbitt:
- updated CLI to automatically add the user's ip address to configs.
- refactored file creation to one function and switched all config files to be defaults rather than keeping some in version control


billbitt:
@daovist I'm adding you as a reviewer for learning.  Lmk if you have any questions

PR #591, opened 2018-09-02
Update babel packages to match major, fix dependency issues

skhameneh:
Tested with yarn `1.7.0` and npm `5.6.0` (included with Node v8.11)

I'm not including a yarn.lock (yet) since we aren't using it universally.


skhameneh:
@nikooo777 the yarn.lock was added by mistake.
Why aren't the servers on LTS?

Issue #184, opened 2017-10-03
Only use the canonical / permanent URL


kauffj:
For SEO purposes, each discrete piece of content should only have a single canonical URL. Currently, in spee.ch, a piece of content can exist on up to 4 URLs:

- /file
- /claim/file
- /channel/file
- /channel:claim/file

All 4 of these pages should include a `<link rel="canonical">` tag to a single URL. That URL should be the one described by lbryio/lbry#958

More on these tags here: https://en.wikipedia.org/wiki/Canonical_link_element

Additionally, we ought to consider if an outright redirect (or JS based updating of the location bar) so that users are encouraged to use one URL over another. In particular, if content is published via a channel, we probably want to encourage users to use channel based URLs.

### Acceptance Criteria
1.
2.
3.

### Definition of Done
- [ ]  Tested against acceptance criteria
- [ ] Tested against the assumptions of the user story
- [ ] The project builds without errors
- [ ] Unit tests are written and passing
- [ ] Tests on devices/browsers listed in the issue have passed
- [ ] QA performed & issues resolved
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented
- [ ] Documentation updated
- [ ] Peer Code Review performed

kauffj:
The URLs described by lbryio/lbry#958 are the _only_ URLs that the app should directly send a user to and are the URLs that should be shown in all embed/share codes.

Additional URLs can resolve, but should include the canonical URL tag.

daovist:
Most of this is done and live on spee.ch. Publishes are not yet using this feature. That will be part of the update-abandon PR upon which this ticket will be closed.

PR #593, opened 2018-09-04
Add chainquery dependencies for Spee.ch, does not include migrations

skhameneh:
Ready for review, uses chainquery first then falls back to local; this is needed for my next PRs for trending content and metrics.


server/chainquery/index.js
```diff
@@ -0,0 +1,80 @@
+const Sequelize = require('sequelize');
+const logger = require('winston');
+
+import abnormalClaimTable from './tables/abnormalClaimTable';
+import addressTable from './tables/addressTable';
+import applicationStatusTable from './tables/applicationStatusTable';
```
tiger5226: This table is used for tracking background job statuses internal to chainquery. Most likely not useful for speech.

server/chainquery/index.js
```diff
@@ -0,0 +1,80 @@
+const Sequelize = require('sequelize');
+const logger = require('winston');
+
+import abnormalClaimTable from './tables/abnormalClaimTable';
```
tiger5226: Is this useful to you? This is for claims that were not made through the app. Claims that are not structured according to the lbry protocol.

server/chainquery/index.js
```diff
@@ -0,0 +1,80 @@
...
+import addressTable from './tables/addressTable';
+import applicationStatusTable from './tables/applicationStatusTable';
+import blockTable from './tables/blockTable';
+import claimTable from './tables/claimTable';
+import gorpMigrationsTable from './tables/gorpMigrationsTable';
+import inputTable from './tables/inputTable';
+import jobStatusTable from './tables/jobStatusTable';
```
tiger5226: same thing. this is an internal table of no importance to speech.

server/chainquery/models/AbnormalClaimModel.js
```diff
@@ -0,0 +1,65 @@
...
+    value_as_hex: {
+      type: TEXT,
+      set() { },
+    },
+    value_as_json: {
+      type: TEXT,
+      set() { },
+    },
+    created_at: {
+      type: INTEGER,
```
tiger5226: These are timestamps, is it ok to create it as an integer?

server/chainquery/models/ClaimModel.js
```diff
@@ -0,0 +1,177 @@
...
+      case 'image/jpg':
+        return 'jpg';
+      case 'image/png':
+        return 'png';
+      case 'image/gif':
+        return 'gif';
+      case 'video/mp4':
+        return 'mp4';
+      default:
+        logger.debug('setting unknown file type as file extension jpg');
```
tiger5226: Is jpg really a good default? Why not blank or unknown?

tiger5226: Also how do you handle nulls? A channel claim is not going to have any of the meta data fields. They will be null in the DB.

server/chainquery/models/ClaimModel.js
```diff
@@ -0,0 +1,177 @@
...
+    description: {
+      type: STRING,
+      set() { },
+    },
+    content_type: {
+      type: STRING,
+      set() { },
+    },
+    is_nsfw: {
+      type: INTEGER,
```
tiger5226: This is a boolean that is a `TINYINT(1)` in the database.

server/chainquery/models/ClaimModel.js
```diff
@@ -0,0 +1,177 @@
...
+    fee_currency: {
+      type: STRING,
+      set() { },
+    },
+    is_filtered: {
+      type: INTEGER,
+      set() { },
+    },
+    bid_state: {
+      type: STRING,
```
tiger5226: This is also an enumeration. I see that you set an enumeration above. Should this be listed as well? `Active,Expired,Controlling,Spent,Accepted` are the enumerations.

server/chainquery/models/InputModel.js
```diff
@@ -0,0 +1,77 @@
...
+    transaction_hash: {
+      type: STRING,
+      set() { },
+    },
+    input_address_id: {
+      type: INTEGER,
+      set() { },
+    },
+    is_coinbase: {
+      type: INTEGER,
```
tiger5226: boolean

server/chainquery/models/InputModel.js
```diff
@@ -0,0 +1,77 @@
...
+    coinbase: {
+      type: STRING,
+      set() { },
+    },
+    prevout_hash: {
+      type: STRING,
+      set() { },
+    },
+    prevout_n: {
+      type: INTEGER,
```
tiger5226: this is unsigned. Does it matter that you treat it as just an integer?

server/chainquery/models/OutputModel.js
```diff
@@ -0,0 +1,77 @@
...
+    },
+    script_pub_key_hex: {
+      type: TEXT,
+      set() { },
+    },
+    required_signatures: {
+      type: INTEGER,
+      set() { },
+    },
+    address_list: {
```
tiger5226: This is a JSON array. Not sure if you can specify that here or not.

server/chainquery/models/OutputModel.js
```diff
@@ -0,0 +1,77 @@
...
+      type: TEXT,
+      set() { },
+    },
+    is_spent: {
+      type: INTEGER,
```
tiger5226: boolean

tiger5226:
Generally, I think there are some tables here that are just not needed for speech and are internal tables for Chainquery. They also don't fall under the major version promise of Chainquery. Those are subject to change in the future. Also there are fields that appear to not have mapped over properly. It may be intentional but I flagged them just in case.

server/chainquery/models/ClaimModel.js
skhameneh: This is the same implementation as existing code in Spee.ch; it's implemented as a generated field.
I may need to revisit this for reusability.

server/chainquery/models/AbnormalClaimModel.js
skhameneh: This will work for now, I need to verify that `Sequelize.DATE(6)` is fully compatible first.
There's a few things like `INT(1)` that I'm using `INTEGER` for as well that could probably be switched to `Sequelize.BOOLEAN`.  I know there's a number of `BIGINT` as well that I need to adjust.

skhameneh:
@tiger5226 which fall under `major version promise`?  I'm aiming to keep this independent of the implementations on Spee.ch

I may need to rethink some of generated fields.

server/chainquery/models/InputModel.js
skhameneh: Ha, alright!  I didn't realize `TINYINT(1)` is exactly the same as `BOOL`
https://dev.mysql.com/doc/refman/5.7/en/numeric-type-overview.html
> **BOOL, BOOLEAN**
> These types are synonyms for TINYINT(1). A value of zero is considered false. Nonzero values are considered true

tiger5226:
It is more about which ones don't. The below tables are used internally by chainquery:
gorpMigrationsTable
applicationStatusTable
jobStatusTable

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
+const logger = require('winston');
+
+const returnShortId = (claimsArray, longId) => {
```
tiger5226: Is this the canonical url?

tiger5226: If so this doesn't seem right. You want the claim_name included here. Also the claim_id is a hash not based on the height. Ignore if this is not the case. @kauffj has a preference for how we do these shortened URLs.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+        throw new Error('No claim(s) found with that claim name');
+      }
+
+      return returnShortId(result, claimId);
+    });
+  },
+
+  getAllChannelClaims: async (channelClaimId) => {
+    logger.debug(`claim.getAllChannelClaims for ${channelClaimId}`);
+    return await table.findAll({
```
tiger5226: you will want to select specific pieces of data. Otherwise you will get all of it no? The metadata is also in this table as columns, json, and as hex. You probably only want 1 of these versions for your query.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+          logger.warn(`${result.length} records found for "${claimName}" in channel "${channelClaimId}"`);
+          return result[0].claim_id;
+      }
+    });
+  },
+
+  validateLongClaimId: async (name, claimId) => {
+    return await table.findOne({
+      where: {
+        name,
```
tiger5226: why is name here?

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+      return claimId;
+    });
+  },
+
+  getLongClaimIdFromShortClaimId: async (name, shortId) => {
+    return await table.findAll({
+      where: {
+        name,
+        claim_id: {
+          [sequelize.Op.like]: `${shortId}%`,
```
tiger5226: This is a very expensive call. Maybe if you really need to use this, I add this feature to Chainquery so it can simply be matched against.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+
+      return channelClaimsArray;
+    })
+  },
+
+  getClaimIdByLongChannelId: async (channelClaimId, claimName) => {
+    logger.debug(`finding claim id for claim ${claimName} from channel ${channelClaimId}`);
+    return await table.findAll({
+      where: { name: claimName, publisher_id: channelClaimId },
+      order: [['id', 'ASC']],
```
tiger5226: any reason for ordering? `id` doesn't mean much so it may be wasteful to order the results, then again I think this is the default ordering for mysql anyway.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+    })
+    .then(result => {
+      switch (result.length) {
+        case 0:
+          return null;
+        case 1:
+          return result[0].claim_id;
+        default:
+          // Does this actually happen??? (from converted code)
```
tiger5226: This is a good question actually. Technically, I believe this can happen in the blockchain. In the end every claim has a bid state. So you can always have a claim for a name and you can have 100 of them. Only one can be in the `Controlling` state though.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+
+      return result[0].claim_id;
+    });
+  },
+
+  getTopFreeClaimIdByClaimName: async (name) => {
+    return await table.findAll({
+      // TODO: Limit 1
+      where: { name },
+      order: [['effective_amount', 'DESC'], ['height', 'ASC']],
```
tiger5226: So not sure we should use this. The `bid_state` = `Controlling` is better. This information comes directly from the blockchain.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+      return table.queries.getLongClaimIdFromShortClaimId(claimName, claimId);
+    } else {
+      return table.queries.getTopFreeClaimIdByClaimName(claimName);
+    }
+  },
+
+  resolveClaim: async (name, claimId) => {
+    logger.debug(`Claim.resolveClaim: ${name} ${claimId}`);
+    return table.findAll({
+      where: { name, claim_id: claimId },
```
tiger5226: not sure why name is here like this.

server/chainquery/queries/claimQueries.js
```diff
@@ -0,0 +1,184 @@
...
+
+      return claimArray[0];
+    });
+  },
+
+  getOutpoint: async (name, claimId) => {
+    logger.debug(`finding outpoint for ${name}#${claimId}`);
+
+    return await table.findAll({
+      where     : { name, claim_id: claimId },
```
tiger5226: I think I get it. You are not specifying name because they are identical. However, the claim table is unique on `claim_id`. So you do not need to ever you both. There is no benefit that I can think of. If there is let me know.

server/controllers/api/claim/data/index.js
```diff
@@ -11,7 +13,14 @@ const claimData = ({ ip, originalUrl, body, params }, res) => {
   const claimName = params.claimName;
   let claimId = params.claimId;
   if (claimId === 'none') claimId = null;
-  db.Claim.resolveClaim(claimName, claimId)
+  chainquery.claim.queries.resolveClaim(claimName, claimId)
+    .then(claimInfo => {
+      if (!claimInfo) {
+        // Not found remote, try local
+        return db.Claim.resolveClaim(claimName, claimId)
```
tiger5226: I like the backup for now. Good for resilience for intial cutover.

server/chainquery/queries/claimQueries.js
skhameneh: The old code validates against the name as well as the claim ID.
I kept the existing functionality - I might remove this.

server/chainquery/queries/claimQueries.js
skhameneh: Good call

server/chainquery/queries/claimQueries.js
skhameneh: Retained from the old query, I'll remove.

server/controllers/api/claim/data/index.js
skhameneh: It's also, unfortunately, required since we don't wait for `chainquery` to have the claim before rendering the content page.  We have a number of nuances that require the fallback and prevent us from fully dropping tables.

server/routes/assets/index.js
```diff
@@ -1,7 +1,11 @@
...
 
-module.exports = (app) => {
-  app.get('/:identifier/:claim', serveByIdentifierAndClaim);
-  app.get('/:claim', serveByClaim);
+// TODO: Adjust build & sources to use import/export everywhere
+const Actions = require('@actions').default;
+const Sagas = require('@sagas').default;
+
+module.exports = {
+  '/:identifier/:claim': { controller: serveByIdentifierAndClaim, action: Actions.onHandleShowPageUri, saga: Sagas.handleShowPageUri },
```
skhameneh: @daovist Take a look at the action/saga binding I've added.
You can also use `httpContext.set('key', value)` as well as `httpContext.get('key')` in controllers.  I'm using this to communicate route information through the stack programmatically.

daovist: Is this for SSR?

skhameneh: Some of it is

server/chainquery/models/AddressModel.js
```diff
@@ -17,15 +22,15 @@ export default (sequelize, {
       set() { },
     },
     first_seen: {
-      type: INTEGER,
+      type: DATE(6),
```
tiger5226: Not sure what `Date(6)` is but the db stores a `DATETIME`

skhameneh: `DATE(6)` is the equivalent

server/chainquery/models/ClaimModel.js
```diff
@@ -145,19 +150,19 @@ export default (sequelize, {
       set() { },
     },
     is_filtered: {
-      type: INTEGER,
+      type: BOOLEAN,
```
tiger5226: This is an old column. I need to delete this. Not sure why it is in the schema. I will create an issue for it.

server/chainquery/queries/claimQueries.js
skhameneh: It's an old query, these can be adjusted over time

server/chainquery/index.js
skhameneh: Not at the moment, retained for full functionality if this is to be split out into a separate lib

server/chainquery/models/OutputModel.js
skhameneh: Can't, not supported unless using a virtual field.  I might add a custom getter method.

server/chainquery/queries/claimQueries.js
kauffj: This is the existing logic, I believe. I think there are some issues with it, but we're going to handle it through https://github.com/lbryio/spee.ch/issues/184

Issue #554, opened 2018-08-06
Allow Deleting (claim abandoning)


kauffj:
To follow #482

Issue #482, opened 2018-06-15
Allow Editing


kauffj:
Let me change the metadata or even the file itself.

skhameneh:
After talking to Bill, need to clarify if edits will update an existing claim or just create a brand new claim with old data pre-populated (as you and I previously discussed)

tzarebczan:
Preferably it would be an update to the claim. The claim id is part of the short URL for my content, so unless you create a redirect, you'd want this to stay the same. The old claim would still need to be abandoned if we decided to make a new claim  - so that's 2 transactions instead of 1. 

I would assume this would only be possible on authenticated channels, unless we introduce a new feature where we give the user an option to specify an "edit password" when publishing.

PR #595, opened 2018-09-06
 allow claims to be updated and abandoned 

daovist:
#482 
#554


client/src/containers/PublishDetails/index.js
```diff
@@ -1,16 +1,29 @@
 import { connect } from 'react-redux';
-import { clearFile, startPublish } from '../../actions/publish';
+import { clearFile, startPublish, abandonClaim } from '../../actions/publish';
 import View from './view';
 
-const mapStateToProps = ({ channel, publish }) => {
+const mapStateToProps = ({ show, publish }) => {
+  // select request info
+  const requestId = show.request.id;
```
skhameneh: You've repeated this code a few times for getting the `asset`, I'd suggest creating a util method

client/src/pages/EditPage/index.js
```diff
@@ -0,0 +1,31 @@
...
+  const { show } = props;
+  // select request info
+  const requestId = show.request.id;
+  // select asset info
+  let asset;
+  const request = show.requestList[requestId] || null;
+  const assetList = show.assetList;
+  if (request && assetList) {
+    const assetKey = request.key;  // note: just store this in the request
+    asset = assetList[assetKey] || null;
```
skhameneh: See above comment about adding a util

client/src/reducers/show.js
```diff
@@ -55,6 +55,33 @@ export default function (state = initialState, action) {
...
+    case actions.ASSET_REMOVE:
+      const claim = action.data;
+      const newAssetList = state.assetList;
+      delete newAssetList[`a#${claim.name}#${claim.claimId}`];
+
+      const channelId = `c#${claim.channelName}#${claim.certificateId}`;
+      const channelClaims = state.channelList[channelId].claimsData.claims;
+      const newClaimsData = channelClaims.filter(c => c.claimId !== claim.claimId);
+
+      return Object.assign({}, state, {
```
skhameneh: Please use spread operators to clean this up.

e.g.:
```
{
  ...state,
  assetList: newAssetList,
  channelList: {
    ...state.channelList,
    [channelId]: {}
  }
}
```

client/src/reducers/show.js
```diff
@@ -55,6 +55,33 @@ export default function (state = initialState, action) {
...
+        channelList: Object.assign({}, state.channelList, {
+          [channelId]: Object.assign({}, state.channelList[channelId], {
+            claimsData: Object.assign({}, state.channelList[channelId].claimsData, {
+              claims: newClaimsData,
+            }),
+          }),
+        }),
+      });
+    case actions.ASSET_UPDATE_CLAIMDATA:
+      return Object.assign({}, state, {
```
skhameneh: Same as above, use spread operators

client/src/sagas/abandon.js
```diff
@@ -0,0 +1,30 @@
+import { call, put, takeLatest } from 'redux-saga/effects';
+import * as actions from '../constants/publish_action_types';
+import * as publishStates from '../constants/publish_claim_states';
+import { updatePublishStatus, clearFile } from '../actions/publish';
+import { removeAsset } from '../actions/show';
+import { doAbandonClaim } from '../api/assetApi';
+
+function * abandonClaim (action) {
+  const { claimData, history } = action.data;
+  const {claimId} = claimData;
```
skhameneh: Should use consistent formatting (I know, the existing code is inconsistent)

client/src/sagas/abandon.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const confirm = window.confirm('Are you sure you want to abandon this claim? This action cannot be undone.');
```
skhameneh: This works for now, but we should have some common prompt to match Spee.ch.

client/src/sagas/publish.js
```diff
@@ -5,44 +5,77 @@ import { updateError, updatePublishStatus, clearFile } from '../actions/publish'
...
-  let publishMetadata = createPublishMetadata(claim, file, metadata, publishInChannel, selectedChannel);
-  if (thumbnail) {
-    // add thumbnail to publish metadata
-    publishMetadata['thumbnail'] = createThumbnailUrl(thumbnailChannel, thumbnailChannelId,  claim, host);
+
+  let publishMetadata, publishFormData, publishChannel;
+  if (isUpdate) {
+    publishMetadata = createPublishMetadata(asset.name, {type: asset.claimData.contentType}, metadata, publishInChannel, selectedChannel);
+    publishMetadata['channelName'] = asset.claimData.channelName;
+    if (thumbnail) {
```
skhameneh: This logic exists in both conditions, can you move it out?

client/src/sagas/publish.js
```diff
@@ -5,44 +5,77 @@ import { updateError, updatePublishStatus, clearFile } from '../actions/publish'
...
+      // add thumbnail to publish metadata
+      publishMetadata['thumbnail'] = createThumbnailUrl(thumbnailChannel, thumbnailChannelId, claim, host);
+    }
+    // create form data for main publish
+    publishFormData = createPublishFormData(file, thumbnail, publishMetadata);
+    // make the publish request
+    publishChannel = yield call(makePublishRequestChannel, publishFormData, true);
```
skhameneh: Can move this out as `publishChannel = yield call(makePublishRequestChannel, publishFormData, isUpdate);`

server/controllers/api/claim/update/updatePublishParams.js
```diff
@@ -0,0 +1,56 @@
+const path = require('path');
+const validateFileTypeAndSize = require('../publish/validateFileTypeAndSize.js');
+
+const parseUpdateFile = ({file}) => {
+  // make sure a file was provided
+  if (!file) {
```
skhameneh: You have this validation in more than one place, please combine

client/src/actions/publish.js
```diff
@@ -14,6 +14,12 @@ export function clearFile () {
   };
 }
 
+export function setUpdateTrue () {
```
kauffj: Is this necessary? My inclination is that there must be another state value corresponding to what is being edited (or even however you ended up on the edit page) that should allow you to know this without an additional explicit value.

client/src/api/assetApi.js
```diff
@@ -37,3 +37,15 @@ export function checkClaimAvailability (claim) {
   const url = `/api/claim/availability/${claim}`;
   return Request(url);
 }
+
+export function doAbandonClaim (claimId) {
+  const params = {
+    method : 'POST',
+    body   : JSON.stringify({claimId}),
+    headers: new Headers({
```
kauffj: DRY

client/src/channels/publish.js
```diff
@@ -1,8 +1,8 @@
 import {buffers, END, eventChannel} from 'redux-saga';
 
-export const makePublishRequestChannel = (fd) => {
+export const makePublishRequestChannel = (fd, isUpdate) => {
   return eventChannel(emitter => {
-    const uri = '/api/claim/publish';
+    const uri = `/api/claim/${isUpdate ? 'update' : 'publish'}`;
     const xhr = new XMLHttpRequest();
```
kauffj: I know you didn't add this, but you should know we can just use the Fetch API.

client/src/constants/publish_claim_states.js
```diff
@@ -3,3 +3,4 @@ export const LOADING = 'LOADING';
 export const PUBLISHING = 'PUBLISHING';
 export const SUCCESS = 'SUCCESS';
```
kauffj: should be `SUCCEEDED` fwiw

client/src/containers/AssetDisplay/view.jsx
```diff
@@ -2,14 +2,49 @@ import React from 'react';
 import ProgressBar from '@components/ProgressBar';
 import { LOCAL_CHECK, UNAVAILABLE, ERROR, AVAILABLE } from '../../constants/asset_display_states';
 
+class AvailableContent extends React.Component {
```
kauffj: good refactor!

client/src/containers/AssetDisplay/view.jsx
```diff
@@ -2,14 +2,49 @@ import React from 'react';
...
 class AssetDisplay extends React.Component {
   componentDidMount () {
     const { asset: { claimData: { name, claimId } } } = this.props;
     this.props.onFileRequest(name, claimId);
   }
   render () {
-    const { status, error, asset: { claimData: { name, claimId, contentType, fileExt, thumbnail } } } = this.props;
-    const sourceUrl = `/${claimId}/${name}.${fileExt}`;
+    const { status, error, asset: { name, claimData: { claimId, contentType, fileExt, thumbnail, outpoint } } } = this.props;
+    const sourceUrl = `/${claimId}/${name}.${fileExt}?${outpoint}`;
```
kauffj: if this isn't being fixed before shipping, it needs to be filed, and it's a good practice to link the issue as an inline comment

client/src/containers/AssetTitle/view.jsx
```diff
@@ -1,10 +1,14 @@
...
 import Row from '@components/Row';
 
-const AssetTitle = ({ title }) => {
+const AssetTitle = ({ title, editable, claimId, name }) => {
   return (
     <Row>
-      <h3>{title}</h3>
+      <h3>
+        {title}
+        {editable && (<span> (<Link to={`/edit/${claimId}/${name}`}>edit</Link>)</span>)}
```
kauffj: better to have this out of the h3 ("edit" isn't part of the title for SEO / accessibility purposes, etc.)

client/src/pages/EditPage/view.jsx
```diff
@@ -0,0 +1,42 @@
...
+    // redirect if user does not own this claim
+    if (
+      !myChannel || (
+        asset &&
+        asset.claimsData &&
+        asset.claimsData.channelName &&
+        asset.claimsData.channelName !== myChannel
+      )
+    ) {
+      return (<Redirect to={'/'} />);
```
kauffj: Do we need to redirect to login? Show an error? How would this happen to a user genuinely? 

(one guess: sign in in one tab, then sign in a second tab, and click edit from the first tab)

client/src/pages/HomePage/view.jsx
```diff
@@ -4,6 +4,9 @@ import PageLayout from '@components/PageLayout';
 import PublishTool from '@containers/PublishTool';
 
 class HomePage extends React.Component {
+  componentWillUnmount () {
+    this.props.clearFile();
```
kauffj: Why does this have to happen here? Does this have to happen for any component that includes `<PublishTool />`? If so, this should be avoided.

server/controllers/api/claim/abandon/index.js
```diff
@@ -0,0 +1,44 @@
...
+const db = require('../../../../models');
+const { abandonClaim } = require('../../../../lbrynet');
+const deleteFile = require('../publish/deleteFile.js');
+const authenticateUser = require('../publish/authentication.js');
+
+/*
+  route to abandon a claim through the daemon
+*/
+
+const claimAbandon = async (req, res) => {
```
kauffj: This function looks great! Looking forward to more like this

server/controllers/api/claim/publish/publish.js
```diff
@@ -1,78 +1,50 @@
...
+    const upsertCriteria = {name: publishParams.name, claimId};
+    if (newFile) {
+      fileRecord = await createFileRecordDataAfterPublish(fileName, fileType, publishParams, publishResults);
+    } else {
+      fileRecord = await db.File.findOne({where: {claimId}});
+    }
+
+    const [file, claim] = await Promise.all([
+      db.upsert(db.File, fileRecord, upsertCriteria, 'File'),
+      db.upsert(db.Claim, claimRecord, upsertCriteria, 'Claim'),
```
kauffj: Do these need to be in a transaction if they need to both execute atomically?

server/controllers/api/claim/update/index.js
```diff
@@ -0,0 +1,151 @@
...
+        title      : claimRecord.title,
+        description: claimRecord.description,
+        nsfw       : claimRecord.nsfw,
+        license    : claimRecord.license,
+        language   : 'en',
+        author     : details.title,
+      }, updateMetadata({title, description, nsfw, license}));
+      const publishParams = {
+        name,
+        bid          : 0.01,
```
kauffj: Can be filed, but this is a rather important value that ought to be hoisted all the way to a site-specific config

client/src/actions/publish.js
daovist: I think it would be more complex to do this another way e.g. looking at the router path

client/src/pages/EditPage/view.jsx
daovist: I was only considering a nefarious user. I don't think this matters much; just needs to not render the edit page.

client/src/pages/HomePage/view.jsx
daovist: Nice catch/good point! Current speech only clears the file for file errors, reset, publish success, and cancel publish, so two features using the same `state.publish` were stepping on each others toes

My new approach uses `isUpdate`

When the publish page mounts: if `isUpdate` then `clearFile`

When edit page mounts: if `!isUpdate` then `clearFile, setUpdateTrue, setMetadata`

This way a user can start a publish or update, leave the page, and come back to where they were, like current speech, unless they go to the other (Edit/Publish) page, in which case it starts fresh with publish or update

[code](https://github.com/lbryio/spee.ch/pull/595/commits/7c4323ba648fea1e8e2b361a17654b10ae66ff93)

kauffj: :+1:

server/controllers/api/claim/update/index.js
daovist: #609

skhameneh:
Is this ready to test?

client/src/containers/Dropzone/index.js
```diff
@@ -1,13 +1,28 @@
...
-  };
+const { assetDefaults: { thumbnail: defaultThumbnail } } = siteConfig;
+
+const mapStateToProps = ({ show, publish: { file, thumbnail, fileError, isUpdate } }) => {
+  const obj = { file, thumbnail, fileError, isUpdate };
+  let asset, name, claimId, fileExt, outpoint, sourceUrl;
+  if (isUpdate) {
+    asset = selectAsset(show);
+    if (asset) {
+      obj.fileExt = asset.claimData.contentType.split('/')[1];
```
skhameneh: store `asset.claimData` in a local var since it's use many times

client/src/containers/Dropzone/view.jsx
```diff
@@ -81,53 +81,70 @@ class Dropzone extends React.Component {
...
+              onDragEnd={this.handleDragEnd}
+              onDragEnter={this.handleDragEnter}
+              onDragLeave={this.handleDragLeave}
+              onMouseEnter={this.handleMouseEnter}
+              onMouseLeave={this.handleMouseLeave}
+              onClick={this.handleClick}>
+              {file || isUpdate ? (
+                <div className={'dropzone-preview-wrapper'}>
+                  {file ? (
+                    <DropzonePreviewImage
```
skhameneh: Can use spread operator for switching props, e.g.:
```
<Elem ...(showSmall ? { small: true } : { small: false }) />
```

server/chainquery/bundle.js
```diff
@@ -981,6 +986,24 @@ var claimQueries = (db, table, sequelize) => ({
     });
   },
 
+  resolveClaimInChannel: async (claimName, channelId) => {
+    logger$1.debug(`Claim.resolveClaimByNames: ${claimName} in ${channelId}`);
+    return table.findAll({
```
skhameneh: Should we order these?

server/controllers/api/claim/abandon/index.js
```diff
@@ -0,0 +1,44 @@
+const logger = require('winston');
+const db = require('../../../../models');
```
skhameneh: use the aliases

server/controllers/api/claim/abandon/index.js
```diff
@@ -0,0 +1,44 @@
...
+const { abandonClaim } = require('../../../../lbrynet');
```
skhameneh: use aliases

server/controllers/api/claim/publish/index.js
```diff
@@ -17,6 +17,9 @@ const parsePublishApiRequestBody = require('./parsePublishApiRequestBody.js');
 const parsePublishApiRequestFiles = require('./parsePublishApiRequestFiles.js');
 const authenticateUser = require('./authentication.js');
 
+const chainquery = require('chainquery');
+const createCanonicalLink = require('../../../../../utils/createCanonicalLink');
```
skhameneh: Use aliases

server/controllers/api/claim/update/index.js
```diff
@@ -0,0 +1,199 @@
+const logger = require('winston');
+const db = require('../../../../models');
```
skhameneh: Use aliases

server/utils/fetchClaimData.js
```diff
@@ -0,0 +1,19 @@
...
+const fetchClaimData = async (params) => {
+  let { claimId, claimName: name } = params;
+  if (claimId === 'none') claimId = null;
+
+  const [cq, local] = await Promise.all([
+    chainquery.claim.queries.resolveClaim(name, claimId).then(res => res.dataValues).catch(() => {}),
+    db.Claim.resolveClaim(name, claimId).catch(() => {}),
+  ]);
+
+  if (!cq && !local) return null;
```
skhameneh: I'd much prefer to avoid single line `if` logic, I find it reduced legibility.

server/chainquery/bundle.js
daovist: This is for a single claim. I need to do a lookup by channel_id. This can be more generalized in the near future.

PR #602, opened 2018-09-17
use publish params for channel data

daovist:


server/controllers/api/claim/publish/publish.js
```diff
@@ -7,34 +7,14 @@ const deleteFile = require('./deleteFile.js');
...
+    let publishResults;
+    const certificateId = publishParams.channel_id || null;
+    const channelName = publishParams.channel_name || null;
     // publish the file
     return publishClaim(publishParams)
       .then(tx => {
         logger.info(`Successfully published ${publishParams.name} ${fileName}`, tx);
         publishResults = tx;
-        // get the channel information
-        if (publishParams.channel_name) {
```
skhameneh: This removes validation, can be dangerous

PR #614, opened 2018-10-02
Removed errors from README.md

DhruvamSharma:


tzarebczan:
@DhruvamSharma, thanks for the edits! Can we send you some [LBC in appreciation](https://lbry.io/faq/appreciation)?

PR #638, opened 2018-10-18
filter claims by bid_state: 'Controlling'

daovist:


server/chainquery/queries/claimQueries.js
```diff
@@ -66,7 +66,7 @@ export default (db, table, sequelize) => ({
   getAllChannelClaims: async (channelClaimId) => {
     logger.debug(`claim.getAllChannelClaims for ${channelClaimId}`);
     return await table.findAll({
-      where: { publisher_id: channelClaimId },
+      where: { publisher_id: channelClaimId, bid_state: 'Controlling' },
```
skhameneh: Do something like:
```getAllChannelClaims: async (channelClaimId, bidState = 'Controlling') => {```

```where: bidState === false ? { publisher_id: channelClaimId } : { publisher_id: channelClaimId, bid_state: 'Controlling' },```

Comment (skhameneh) [reply server/chainquery/queries/claimQueries.js]: This way all can be easily queried by disabling the default filter

File: server/chainquery/queries/claimQueries.js [hunk 2]
```diff
@@ -63,19 +63,22 @@ export default (db, table, sequelize) => ({
     });
   },
 
-  getAllChannelClaims: async (channelClaimId) => {
+  getAllChannelClaims: async (channelClaimId, bidState = 'Controlling') => {
```
tzarebczan: actually, I think this should be either Controlling or Active.

tzarebczan: @tiger5226 can you confirm?

tiger5226: Well they are different bid states. `Controlling` means the claim has won the bid for that name, while `Active` means it has lost. So the question is do they want all claims, regardless of bid state? Then yes, both should be considered. There is also `Accepted`, `Spent` too.

tzarebczan: I didn't look at the overall code, but judging by the variable name it's looking it up by the claim ID, and not the claim name... So I don't see why we'd want only controlling claims.

tiger5226: that's a good point. Basically it will return null if it is not controlling. Seems like this check should be somewhere else more descriptive.

daovist: As of #668 it's getting Controlling, Active, and Spent:

```
    const whereClause = bidState || {
      [sequelize.Op.or]: [
        { bid_state: 'Controlling' },
        { bid_state: 'Active' },
        { bid_state: 'Accepted' },
      ],
    };
```

PR #641, opened 2018-10-19
refactor getClaimIdAndServeAsset using async/await

daovist:
I refactored this function using async/await to debug multisite content blocking. That princess is in another castle but this is a step towards a cleaner code base.

### DIFF
--- a/server/controllers/assets/utils/getClaimIdAndServeAsset.js
+++ b/server/controllers/assets/utils/getClaimIdAndServeAsset.js
@@ -17,81 +17,63 @@ const CONTENT_UNAVAILABLE = 'CONTENT_UNAVAILABLE';
 
 const { publishing: { serveOnlyApproved, approvedChannels } } = require('@config/siteConfig');
 
-const getClaimIdAndServeAsset = (channelName, channelClaimId, claimName, claimId, originalUrl, ip, res) => {
-  getClaimId(channelName, channelClaimId, claimName, claimId)
-    .then(fullClaimId => {
-      claimId = fullClaimId;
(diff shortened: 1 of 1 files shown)


server/controllers/assets/utils/getClaimIdAndServeAsset.js
```diff
@@ -17,81 +17,63 @@ const CONTENT_UNAVAILABLE = 'CONTENT_UNAVAILABLE';
...
+    await db.Blocked.isNotBlocked(claim.outpoint);
+
+    const fileRecord = await db.File.findOne({ where: { name, claimId } });
+    if (!fileRecord) {
+      throw NO_FILE;
+    }
+
+    serveFile(fileRecord.dataValues, res);
+  } catch (error) {
+    if (error === NO_CLAIM) {
```
skhameneh: Use a switich - or at least an `else if` for each subsequent.

server/controllers/assets/utils/getClaimIdAndServeAsset.js
```diff
@@ -17,81 +17,63 @@ const CONTENT_UNAVAILABLE = 'CONTENT_UNAVAILABLE';
...
-      return db.Blocked.isNotBlocked(claim.dataValues.outpoint);
-    })
-    .then(() => {
-      return db.File.findOne({
-        where: {
-          claimId,
-          name: claimName,
-        },
+    let _claim;
+    _claim = await chainquery.claim.queries.resolveClaim(name, claimId);
```
skhameneh: Can be in same line as above

PR #655, opened 2018-10-24
Add zScore, pValue, specials, trending content

skhameneh:
Mostly done, major logic is complete, will be adding some UI polish.
This is mostly PR ready


client/src/pages/ShowChannel/view.jsx
```diff
@@ -7,6 +7,9 @@ import Row from '@components/Row';
 
 class ShowChannel extends React.Component {
   render () {
+    console.log({
```
daovist: remove console log

server/controllers/api/special/claims/index.js
```diff
@@ -0,0 +1,42 @@
...
+
+*/
+
+const channelClaims = async ({ ip, originalUrl, body, params }, res) => {
+  const {
+    name,
+    page,
+  } = params;
+
+  if(name == 'trending') {
```
daovist: should be triple equals?

skhameneh: Yes. I was debugging and forgot to change it back :)

PR #664, opened 2018-10-29
Add views to claim pages for logged in channels

skhameneh:


server/controllers/api/claim/views/index.js
```diff
@@ -0,0 +1,29 @@
...
+const db = require('server/models');
+
+/*
+
+  route to return data for a claim
+
+*/
+
+const claimViews = async ({ ip, originalUrl, body, params }, res) => {
+  const claimName = params.claimName;
```
daovist: `claimName` isn't used here

PR #719, opened 2018-11-10
Redesign

seanyesmunt:
Can be seen at https://sean.lbry.tech.

Spent some more time on mobile, I think that is looking really good. I'm not sure about the show page. I think that can be done better.

cc @kauffj


deployment-config.json
```diff
@@ -0,0 +1,10 @@
+{
```
daovist: this shouldn't be here

PR #747, opened 2018-11-15
moves sensitive keys to gitignored folder

jessopb:
Old behavior: siteConfig had some sensitive passwords.
New behavior: passwords are stored in ./config which is gitignored.

### DIFF
--- a/cli/configure.js
+++ b/cli/configure.js
@@ -3,6 +3,7 @@ const fs = require('fs');
 const Path = require('path');
 const axios = require('axios');
 const ip = require('ip');
+const pwGenerator = require('generate-password');
 
 const mysqlQuestions = require(Path.resolve(__dirname, 'questions/mysqlQuestions.js'));
 const siteQuestions = require(Path.resolve(__dirname, 'questions/siteQuestions.js'));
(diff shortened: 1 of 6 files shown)


cli/configure.js
```diff
@@ -12,11 +13,14 @@ let thumbnailChannelDefault = '@thumbnails';
 let thumbnailChannel = '';
 let thumbnailChannelId = '';
 
-const createConfigFile = (fileName, configObject) => {  // siteConfig.json , siteConfig
-  const fileLocation = Path.resolve(__dirname, `../site/config/${fileName}`);
+const createConfigFile = (fileName, configObject, topSecret) => {  // siteConfig.json , siteConfig
+  const fileLocation = topSecret
+    ? Path.resolve(__dirname, `../config/${fileName}`)
```
skhameneh: When we spoke, I may have said `/config`, which I meant `/site/config` in context.
We don't need a separate directory, but if we use one I would suggest using `site/secrets`.

jessopb: Ok. What you said was literally true: The spee.ch repo's gitignore did indeed cover /config even though it didn't exist. Where would you like the gitignore for the site repo?

PR #755, opened 2018-11-21
fixes blank tiles due to unsupported content

jessopb:
This is a draft of a solution. 

1) What I'd like to do is have a central list of supported type strings ( 'image/jpg' ); passing this list allows for future filtering. Would it be better for supported types be passed or require()d?

2) Also, this function could be renamed or duplicated to GetAllMatchingChannelClaims, GetAllSupportedChannelClaims, or just GetChannelClaims. I don't think it makes sense to duplicate it just to get at currently unsupported content.

### DIFF
--- a/server/chainquery/queries/claimQueries.js
+++ b/server/chainquery/queries/claimQueries.js
@@ -68,15 +68,35 @@ export default (db, table, sequelize) => ({
     });
   },
 
-  getAllChannelClaims: async (channelClaimId, bidState) => {
+  getAllChannelClaims: async (channelClaimId, params) => {
     logger.debug(`claim.getAllChannelClaims for ${channelClaimId}`);
-    const whereClause = bidState || {
-      [sequelize.Op.or]: [
-        { bid_state: 'Controlling' },
-        { bid_state: 'Active' },
-        { bid_state: 'Accepted' },
-      ],
+
+    const defaultWhereClauses = {
+      bid_state:
+      { [sequelize.Op.or]: ['Controlling', 'Active', 'Accepted'] }
     };
+
+    const addWhereClauses = (whereClauses, params) => {
+      /*
+       input params = { col: ['Val', 'Val']}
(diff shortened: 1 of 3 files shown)


server/chainquery/queries/claimQueries.js
```diff
@@ -70,15 +70,25 @@ export default (db, table, sequelize) => ({
...
     logger.debug(`claim.getAllChannelClaims for ${channelClaimId}`);
-    const whereClause = bidState || {
+    const whereBidClause = bidState || {
       [sequelize.Op.or]: [
         { bid_state: 'Controlling' },
         { bid_state: 'Active' },
         { bid_state: 'Accepted' },
       ],
     };
+    const whereContentClause = {
```
skhameneh: This needs to be optional, chainquery functions should not be opinionated on Spee.ch content types.

server/chainquery/queries/claimQueries.js
```diff
@@ -68,15 +68,35 @@ export default (db, table, sequelize) => ({
     });
   },
 
-  getAllChannelClaims: async (channelClaimId, bidState) => {
+  getAllChannelClaims: async (channelClaimId, params) => {
     logger.debug(`claim.getAllChannelClaims for ${channelClaimId}`);
-    const whereClause = bidState || {
```
skhameneh: See here and the original signature of `getAllChannelClaims: async (channelClaimId, bidState)`, this allows all claims to be fetched.  The changes break this pattern and force Spee.ch filtering on code that should be unopinionated.

jessopb: I had read that once you go beyond 1 optional parameter, you're better off passing an object. With bidState already being optional and not passed from the controller, contentTypes would have to be a second optional parameter, meaning the function, I think, would have to be called with ( first, null, third ). Is there a better way to make contentTypes optional?  I'm leading toward a GetMatchingChannelClaims function in chainquery instead. This would pave the way for future db queries.

skhameneh:
Merging to get a fix in ASAP, will make a clean-up task.

PR #794, opened 2018-11-30
proposed readme changes

kauffj:


README.md
```diff
@@ -1,27 +1,23 @@
 # Spee.ch
-Spee.ch is a [NodeJS](https://nodejs.org) React web app that reads and publishes images and videos to and from the [LBRY](https://lbry.io/) blockchain.
+Spee.ch is a [NodeJS](https://nodejs.org) React web app that reads and publishes images, videos and other assets to and from the [LBRY](https://lbry.io/) blockchain.
 
-You can create your own custom version of spee.ch by installing this code base and then creating your own custom components and styles to override the defaults. (More details/guide on how to do that coming soon.)
+Succinctly, via spee.ch, you can have a user-friendly, custom-designed image and video hosting site that is backed by a decentralized network. Via just a set of config files, you can spin your entire site back up including assets, with no backing up necessary.
```
skhameneh: I would suggest omitting some of the superfluous words and fix the grammar (or split up into more, separate, sentences).

Suggestions/comments:
1) No need for `Succinctly`.
2) Mixing clauses (which a comma is missing) with a list greatly reduces legibility.
3) Use nouns before propositions for increased simplicity/legibility (Not a fan of `Via just`)

kauffj: I'm not a fan of my intro wording either :/

Still, the original wording doesn't convey enough about _why_ this is interesting. I'll try again, but I encourage you and @jessopb to try as well.

PR #802, opened 2018-12-03
Fix SCSS overrides for site/custom

skhameneh:
Few things:

1. `sass-loader` will only use aliases if a tilde, `~`, is used.
2. Webpack only allows strings as aliases (not arrays)
3. Aliases/Resolves do not work with a SCSS file in the root path e.g. Can't use `~fileName.scss`.
4. Webpack's resolve pipeline will be used for modules (directory name) even if there is no matching alias.


kauffj:
Nice, thank you!

PR #807, opened 2018-12-04
fixes bug in lbryurl copy

jessopb:
Previous fix using id='short-link' caused copy button to copy contents of another element. This copies as intended.


client/src/containers/AssetInfo/view.jsx
```diff
@@ -27,6 +27,7 @@ class AssetInfo extends React.Component {
         shortId: channelShortId,
       };
       channelCanonicalUrl = `${createCanonicalLink({channel})}`;
+      console.log(channelName)
```
skhameneh: Remove

PR #814, opened 2018-12-07
bugfixes around display of permanentURIs

jessopb:
previous behavior:
 - AssetInfo PermanentURLs for anonymous publishes was broken.
 - Edit claim message "Editing lbry:// had a malformed url.

new behavior:
 - buildURI has been renamed to createPermanentURI to be consistent with lbrynet output.
 - buildURI has been modified to return proper strings for permanent URIs for both anonymous and channeled claims.
 - PublishTools message "editing lbry://abcd" now displays a link text that actually works. 
 - AssetInfo displays and uses proper permanent URIs for both anonymous and channeled claims.
 - @clientutils alias was added to remove confusion about ../../../../../../utils.


utils/createModuleAliases.js
```diff
@@ -34,6 +34,11 @@ module.exports = () => {
   moduleAliases['@config'] = resolve('site/config');
   moduleAliases['@private'] = resolve('site/private');
 
+  // aliases for utils
+  moduleAliases['@globalutils'] = resolve('utils');
```
skhameneh: I would suggest renaming to something like `@rootUtils`

PR #830, opened 2018-12-18
fixes channel claim issue introduced by file extension fix

jessopb:
Previous fix attempted to call split when file_name didn't exist such as on channel claims.
Code now checks for that.
Future not implemented yet: filenames without a . should have '' extension.

### DIFF
--- a/server/utils/getClaimData.js
+++ b/server/utils/getClaimData.js
@@ -22,7 +22,7 @@ module.exports = async (data) => {
   let lbrynetFileExt = null;
   if (!data.fileExt) {
     lbrynetClaimResult = await getClaim(lbrynetUri).catch(() => { return 'invalid URI' });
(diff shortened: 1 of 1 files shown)


server/utils/getClaimData.js
```diff
@@ -22,7 +22,7 @@ module.exports = async (data) => {
   let lbrynetFileExt = null;
   if (!data.fileExt) {
     lbrynetClaimResult = await getClaim(lbrynetUri).catch(() => { return 'invalid URI' });
-    lbrynetFileExt = lbrynetClaimResult && lbrynetClaimResult.file_name.split('.').slice(-1).pop();
+    lbrynetFileExt = lbrynetClaimResult && lbrynetClaimResult.file_name && lbrynetClaimResult.file_name.split('.').slice(-1).pop();
```
skhameneh: I'd like to see more verbose logic, I'm not a fan of this being able to return `null` / `false`.
I'll investigate how we should handle this at a higher level (larger refactor-type issue)

PR #860, opened 2019-01-07
cuts staging from master

jessopb:


server/index.js
```diff
@@ -34,6 +35,9 @@ const {
 
 const { sessionKey } = require('@private/authConfig.json');
 
+// configure.js doesn't handle new keys in config.json files yet. Make sure it doens't break.
+let bLE;
```
skhameneh: I suggest using more descriptive variable names.
Keep names succinct and a descriptive label, even `currentBlocklistEndpoint` would work

PR #878, opened 2019-01-24
fixes bug in edit page due to inconsistent url

jessopb:
A function for populating the channel list details was expecting a shortId was being passed a longId. It was simpler to change it to accept a longId.

### DIFF
--- a/server/controllers/api/channel/claims/getChannelClaims.js
+++ b/server/controllers/api/channel/claims/getChannelClaims.js
@@ -3,30 +3,37 @@ const chainquery = require('chainquery').default;
 const getClaimData = require('server/utils/getClaimData');
 const { returnPaginatedChannelClaims } = require('./channelPagination.js');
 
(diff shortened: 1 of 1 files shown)


server/controllers/api/channel/claims/getChannelClaims.js
```diff
@@ -3,30 +3,37 @@ const chainquery = require('chainquery').default;
...
-    'image/png',
-    'image/gif',
-    'video/mp4',
-  ] };
+const getChannelClaims = async (channelName, channelLongId, page) => {
+  const params = {
+    content_type: ['image/jpeg', 'image/jpg', 'image/png', 'image/gif', 'video/mp4'],
+  };
+
+  let channelShortId = await chainquery.claim.queries.getShortClaimIdFromLongClaimId(
```
jessopb: Line 11 is the main fix, after changing input params from expecting shortId to expecting longId.

PR #896, opened 2019-02-04
adds scss data: option

jessopb:
hotfixes error with scss build from undefined variables.


jessopb:
This would interfere with site custom scss

PR #906, opened 2019-02-09
styling, markdown settings tied to config

jessopb:
links markdown settings to siteConfig
fixes some styling

### DIFF
--- a/cli/defaults/siteConfig.json
+++ b/cli/defaults/siteConfig.json
@@ -42,11 +42,43 @@
   },
   "serving": {
     "markdownSettings": {
-      "skipHtml": true,
-      "privilegedDisallowedTypesDescriptions": ["Image"],
-      "privilegedDisallowedTypesMain": [],
-      "publicDisallowedTypesDescriptions": ["Image"],
-      "publicDisallowedTypesMain": []
+      "skipHtmlMain": true,
+      "escapeHtmlMain": true,
+      "skipHtmlDescriptions": true,
+      "escapeHtmlDescriptions": true,
+      "allowedTypesMain": [],
+      "allowedTypesDescriptions": [],
+      "allowedTypesExample": [
+        "see react-markdown docs",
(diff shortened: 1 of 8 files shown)


client/src/components/FileViewer/index.jsx
```diff
@@ -39,7 +40,7 @@ class FileViewer extends React.Component {
       <div className={'markdown'}>
         {
           this.state.fileLoaded &&
-          <ReactMarkdown className={'markdown-preview'} source={this.state.fileText} skipHtml />
+          <ReactMarkdown className={'markdown-preview'} source={this.state.fileText} skipHtml={skipHtmlMain} disallowedTypes={disallowedTypesMain} escapeHtml={escapeHtmlMain} />
```
skhameneh: This change is **very** dangerous, can you make a list of what is **whitelisted**?

Blacklists are not safe when it comes to script injection.

jessopb: It is now a very long whitelist :)

cli/defaults/siteConfig.json
```diff
@@ -42,11 +42,43 @@
   },
   "serving": {
     "markdownSettings": {
-      "skipHtml": true,
-      "privilegedDisallowedTypesDescriptions": ["Image"],
-      "privilegedDisallowedTypesMain": [],
-      "publicDisallowedTypesDescriptions": ["Image"],
-      "publicDisallowedTypesMain": []
+      "skipHtmlMain": false,
+      "escapeHtmlMain": false,
```
skhameneh: I took a closer look, set this to `true` and make sure it's documented as a dangerous setting.  
I don't want users deploying making mistakes, this needs to be explicitly `true` on all our instances that allow public content.

jessopb:
I made the changes you requested. Also added an ErrorBoundary component to the FileViewer component to prevent it from crashing the app.

PR #912, opened 2019-02-13
enables nsfw updates

jessopb:
* not yet shown in AssetInfo, etc


client/src/containers/PublishMetadataInputs/view.jsx
```diff
@@ -36,9 +36,9 @@ class PublishMetadataInputs extends React.Component {
               description={this.props.description}
               handleInput={this.handleInput}
             />
-            <PublishLicenseInput
```
skhameneh: Why is this commented out?
Does it need to be removed?

jessopb: It didn't appear to be wired up yet, but it does publish. I put it back.

PR #913, opened 2019-02-13
Updated the customise.md

kcseb:
Added the instructions to use npm run build BEFORE npm run start. 

You have to rebuild before starting the spee.ch server again or else it will not take effect.


kcseb:
Closing this due to an error I made, reopened under https://github.com/lbryio/spee.ch/pull/914

PR #917, opened 2019-02-16
adds node-cache and updates handlebars vulnerability

jessopb:


jessopb:
Small bugfix to configs. Not requesting review.

server/utils/cache.js
```diff
@@ -0,0 +1,2 @@
+const NodeCache = require('node-cache');
+module.exports = new NodeCache({ stdTTL: 180, checkperiod: 60 });
```
skhameneh: @jessopb can you demo how you'll be implementing this before merging?

jessopb: The goal to remove localdb claim and file table dependencies. This establishes cache as a near-singleton (though not perfectly due to case insensitivity in package names).
On publish: cache.set( key: claimIdAsKey, value: filePath )
then On any file serve operation, serve `cache.get( claimId)`, else `sdk.file_list(outpoint)[0].path`

The problem to be solved is, given claimId and claimName, How do we get the most up to date file_path from disk? @kauffj mentioned using cache. 

This way, the local speech instance always knows exactly what the brand new publish is for immediate display after publish, and after a handful of seconds chainquery has it in mempool and we're safe.

Why not just sdk.file_list(claimID)? Currently it returns all previous content files in an unsorted order.

Alternate ways to do this: file_list (claimId) and then for each path, find the largest (newest) mtime.

PR #956, opened 2019-03-08
improves preview cards

jessopb:
To preview cards:
adds publish time information
rearranges label components


tzarebczan:
Does this show relative timestamps? We show relative on file tiles and then publish date on claim pages in the app.

jessopb:
Yes, relative timestamps on preview cards based on transactionTime.

PR #973, opened 2019-03-23
cuts release from staging

jessopb:


jessopb:
Travis failure due to autocancellation due to newer version being pushed before completion.

PR #985, opened 2019-04-01
Fixing squished wordwrap lines

D1V3:
Added tabs to markup to clean visual output


D1V3:
@jessopb - looks like just adding tabs should've fixed it

PR #1011, opened 2019-05-30
Added Server Response Error Handling

ProfessorDey:
Presently the spee.ch client will error out if it received an unexpected response, resulting in hangs and a very vague JSON parsing error. This failsafe will catch the known issue of server request size limits causing 413 responses if misconfigured, making things easier to diagnose, as well as catching any other unexpected responses cleanly. Further specific behaviours can be added to ensure administrators spend less time debugging simple configuration issues.

The 413 error response should be fairly self explanatory, sufficient to not need further documentation, though adding an addendum to the README.md would aid other developers have a smooth experience.

### DIFF
--- a/client/src/channels/publish.js
+++ b/client/src/channels/publish.js
@@ -23,14 +23,29 @@ export const makePublishRequestChannel = (fd, isUpdate) => {
     // set state change handler
     xhr.onreadystatechange = () => {
       if (xhr.readyState === 4) {
-        const response = JSON.parse(xhr.response);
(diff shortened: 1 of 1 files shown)


tzarebczan:
Thanks Dey! Sent you some appreciation on Discord.

ProfessorDey:
There you go @jessopb, as per request I fixed the indenting (Was actually an issue where the first and third cases had tabs at the start which threw off the indentation, fixed everything to standard double space, as is most common practice. Also made the requested change for the ready state check.

